PR #428, opened 2024-01-16
chore: rearrange final token validation

sujithsomraaj:


src/crosschain-data/extensions/CoreStateRegistry.sol
```diff
@@ -508,21 +504,17 @@ contract CoreStateRegistry is BaseStateRegistry, ICoreStateRegistry {
             revert Error.ZERO_AMOUNT();
         }
 
-        if (_getVaultAsset(_getSuperform(singleVaultData.superformId)) != finalToken_) {
-            revert Error.INVALID_UPDATE_FINAL_TOKEN();
-        }
-
         /// @dev observe not consuming the third return value
         (singleVaultData.amount, finalState_,) = _updateAmount(
```
0xTimepunk, line 508: Notice finalState_ is updated inside _updateAmount. It should not be removed from the args of _updateAmount as it is used in _updateSingleDeposit (check the returns of this function) 

Suggest leaving that part as is

0xTimepunk (approved):

PR #440, opened 2024-01-18
test: coverage paymentHelper SUP-5257

0xTimepunk:


vikramarun (commented):
LGTM -- one more thing, where do we set the default native prices in setup? Like price of ETH/USDC on Avax to use as a fallback if a feed doesn't exist, etc.

PR #450, opened 2024-01-22
chore: ackGasCost updates

vikramarun:
Optimizing `ackGasCost` based on the chain<>chain connections (WIP--see Slack for options)


script/Abstract.Deploy.Single.s.sol
```diff
@@ -932,13 +932,38 @@ abstract contract AbstractDeploySingle is Script {
...
+
+        uint256 ackCost;
+
+        if (vars.chainId == vars.dstChainId) { // same chain is only minting cost
+            ackCost = 1_000_000;
+        } else {
+            uint256 baseAckCost = 30_000_000; // cheapest cross-chain ack cost
+            uint256 multiplier;
+
+            if (vars.dstChainId == ETH) {
```
0xTimepunk, line 952: The issue with this approach is that this is not scalable for when we add new chains, these values will always be off

PR #486, opened 2024-03-21
feat: Enable Broadcasting in mainnet

0xTimepunk:
This PR reflects changes in the script files to turn on Broadcasting in mainnet. This way SuperPositions can start to be wrapped up in ERC20s and used for a variety of use cases


vikramarun (approved):
lgtm

PR #492, opened 2024-04-04
feat: add deBridge validator and integration test [SUP-6429]

sujithsomraaj:


src/crosschain-liquidity/debridge/DeBridgeValidator.sol
```diff
@@ -0,0 +1,156 @@
...
+        if (deBridgeQuote.externalCall.length > 0) revert();
+
+        /// FIXME: set the new role and add explicit revert message
+        if (
+            superRegistry.getAddressByChainId(keccak256("DEBRIDGE_AUTHORITY"), args_.dstChainId)
+                != _castToAddress(deBridgeQuote.orderAuthorityAddressDst)
+        ) revert();
+
+        /// FIXME: add explicity revert message
+        if (deBridgeQuote.allowedCancelBeneficiarySrc.length > 0) revert();
```
0xTimepunk, line 45: I guess this needs to be allowed?

foundry.toml
```diff
@@ -23,6 +23,7 @@ out = "out"
 script = "script"
 src = "src"
 test = "test"
+memory_limit = 1000000000
```
0xTimepunk, line 26: Remove this

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,247 @@
...
+        DecodedQuote memory deBridgeQuote = _decodeTxData(args_.txData);
+
+        if (
+            superRegistry.getAddressByChainId(keccak256("DEBRIDGE_AUTHORITY"), args_.dstChainId)
+                != deBridgeQuote.orderAuthorityAddressDst
+        ) revert INVALID_DEBRIDGE_AUTHORITY();
+
+        /// @dev 1. chain id calidation
+        /// FIXME: check if this cast is right
+        /// FIXME: check upstream if the srcChain in this context is the block.chainid
```
0xTimepunk, line 84: are these FIXME's old? What was the cast about?

debridge chain ids are equal to block.chainId?

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,247 @@
...
+        if (uint64(deBridgeQuote.dstChainId) != args_.liqDstChainId || args_.liqDataToken != deBridgeQuote.inputToken) {
+            revert Error.INVALID_TXDATA_CHAIN_ID();
+        }
+
+        /// @dev 2. receiver address validation
+        /// @dev allows dst swaps by coupling hashflow with other bridges
```
0xTimepunk, line 90: Nit: unsure what this hashflow comment refers to

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,247 @@
...
+                !(
+                    deBridgeQuote.finalReceiver
+                        == superRegistry.getAddressByChainId(keccak256("CORE_STATE_REGISTRY"), args_.dstChainId)
+                        || hasDstSwap
+                )
+            ) {
+                revert Error.INVALID_TXDATA_RECEIVER();
+            }
+
+            /// @dev if there is a dst swap then the interim token should be the quote of hashflow
```
0xTimepunk, line 110: Nit: should be debridge

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,247 @@
...
+
+    /// @inheritdoc BridgeValidator
+    function decodeDstSwap(bytes calldata /*txData_*/ )
+        external
+        pure
+        override
+        returns (address, /*token_*/ uint256 /*amount_*/ )
+    {
+        /// @dev debridge cannot be used for just swaps
+        revert Error.CANNOT_DECODE_FINAL_SWAP_OUTPUT_TOKEN();
```
0xTimepunk, line 142: I would use a more meaningful error message here - ONLY_SWAPS_DISALLOWED for eg

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,247 @@
...
+    //////////////////////////////////////////////////////////////
+
+    /// @inheritdoc BridgeValidator
+    function validateReceiver(bytes calldata txData_, address receiver) external view override returns (bool) {
+        DecodedQuote memory deBridgeQuote = _decodeTxData(txData_);
+
+        return (receiver == deBridgeQuote.finalReceiver);
+    }
+
+    /// NOTE: check other parameters including: `givePatchAuthoritySrc`
```
0xTimepunk, line 72: Nit: should use @notice instead of note so that natspec picks it up

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,247 @@
...
+        bytes4 selector;
+        address swapOutputToken;
+        address swapRefundAddress;
+        address bridgeTarget;
+        bytes bridgeTxData;
+        bytes permitEnvelope;
+        DlnOrderLib.OrderCreation xChainQuote;
+    }
+
+    /// NOTE: we should only allow the `tradeXChainRFQT` identifier
```
0xTimepunk, line 165: Nit: should use @notice instead of note so that natspec picks it up

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,247 @@
...
+
+        deBridgeQuote.outputToken = _castToAddress(v.xChainQuote.takeTokenAddress);
+        deBridgeQuote.outputAmount = v.xChainQuote.takeAmount;
+        deBridgeQuote.finalReceiver = _castToAddress(v.xChainQuote.receiverDst);
+        deBridgeQuote.dstChainId = v.xChainQuote.takeChainId;
+        deBridgeQuote.orderAuthorityAddressDst = _castToAddress(v.xChainQuote.orderAuthorityAddressDst);
+    }
+
+    /// @dev helps parsing debridge calldata and return the input parameters
+    function parseCallData(bytes calldata callData) public pure returns (bytes calldata) {
```
0xTimepunk, line 231: Any specific reason this is public?

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,247 @@
...
+            (v.xChainQuote,,,, v.permitEnvelope,) = abi.decode(
+                this.parseCallData(v.bridgeTxData), (DlnOrderLib.OrderCreation, uint64, bytes, uint32, bytes, bytes)
+            );
+        } else {
+            revert Error.BLACKLISTED_ROUTE_ID();
+        }
+
+        if (v.swapOutputToken != v.xChainQuote.giveTokenAddress) revert INVALID_BRIDGE_TOKEN();
+
+        if (v.xChainQuote.givePatchAuthoritySrc != address(0) || v.xChainQuote.allowedCancelBeneficiarySrc.length > 0) {
```
0xTimepunk, line 211: I would split the `v.xChainQuote.allowedCancelBeneficiarySrc.length > 0` from here and give it a more meaningful error message, bc INVALID_SRC_DEBRIDGE_AUTHORITY doesn't seem to pertain to it?

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,247 @@
...
+    /// @dev helps parse bytes memory selector
+    function _parseSelectorMem(bytes memory data) internal pure returns (bytes4 selector) {
+        assembly {
+            selector := mload(add(data, 0x20))
+        }
+    }
+
+    /// @dev helps cast bytes to address
+    function _castToAddress(bytes memory address_) internal pure returns (address) {
+        /// FIXME: check if address(uint160(uint256(b))) could be true ??
```
0xTimepunk, line 244: why would we need to perform this check? maybe consider removing this

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,247 @@
...
+        if (v.xChainQuote.externalCall.length > 0) {
+            revert INVALID_EXTRA_CALL_DATA();
+        }
+
+        deBridgeQuote.outputToken = _castToAddress(v.xChainQuote.takeTokenAddress);
+        deBridgeQuote.outputAmount = v.xChainQuote.takeAmount;
+        deBridgeQuote.finalReceiver = _castToAddress(v.xChainQuote.receiverDst);
+        deBridgeQuote.dstChainId = v.xChainQuote.takeChainId;
+        deBridgeQuote.orderAuthorityAddressDst = _castToAddress(v.xChainQuote.orderAuthorityAddressDst);
+    }
```
gpersoon, line 229: maybe also check `allowedTakerDst` ?

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
```diff
@@ -0,0 +1,170 @@
...
+
+            /// @dev if there is a dst swap then the interim token should be the quote of hashflow
+            if (hasDstSwap && (args_.liqDataInterimToken != _castToAddress(deBridgeQuote.takeTokenAddress))) {
+                revert Error.INVALID_INTERIM_TOKEN();
+            }
+        } else {
+            /// @dev if withdrawal, then receiver address must be the receiverAddress
+            if (receiver != args_.receiverAddress) revert Error.INVALID_TXDATA_RECEIVER();
+        }
+    }
```
gpersoon, line 96: Maybe also check:
- givePatchAuthoritySrc
- allowedTakerDst
- takeAmount

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
```diff
@@ -0,0 +1,170 @@
...
+        /// @dev 1. chain id calidation
+        /// FIXME: check if this cast is right
+        /// FIXME: check upstream if the srcChain in this context is the block.chainid
+        if (
+            uint64(deBridgeQuote.takeChainId) != args_.liqDstChainId
+                || args_.liqDataToken != deBridgeQuote.giveTokenAddress
+        ) revert Error.INVALID_TXDATA_CHAIN_ID();
+
+        /// @dev 2. receiver address validation
+        /// @dev allows dst swaps by coupling hashflow with other bridges
```
gpersoon, line 67: is hashflow correct here?

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
```diff
@@ -0,0 +1,170 @@
...
+            if (
+                !(
+                    receiver == superRegistry.getAddressByChainId(keccak256("CORE_STATE_REGISTRY"), args_.dstChainId)
+                        || hasDstSwap
+                )
+            ) {
+                revert Error.INVALID_TXDATA_RECEIVER();
+            }
+
+            /// @dev if there is a dst swap then the interim token should be the quote of hashflow
```
gpersoon, line 88: is hashflow correct here?

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,247 @@
...
+    /// @inheritdoc BridgeValidator
+    function validateTxData(ValidateTxDataArgs calldata args_) external view override returns (bool hasDstSwap) {
+        DecodedQuote memory deBridgeQuote = _decodeTxData(args_.txData);
+
+        if (
+            superRegistry.getAddressByChainId(keccak256("DEBRIDGE_AUTHORITY"), args_.dstChainId)
+                != deBridgeQuote.orderAuthorityAddressDst
+        ) revert INVALID_DEBRIDGE_AUTHORITY();
+
+        /// @dev 1. chain id calidation
```
gpersoon, line 85: calidation typo?

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
```diff
@@ -0,0 +1,170 @@
...
+
+        if (
+            superRegistry.getAddressByChainId(keccak256("DEBRIDGE_AUTHORITY"), args_.dstChainId)
+                != _castToAddress(deBridgeQuote.orderAuthorityAddressDst)
+        ) revert INVALID_DEBRIDGE_AUTHORITY();
+
+        /// FIXME: add explicity revert message
+        if (deBridgeQuote.allowedCancelBeneficiarySrc.length > 0) revert();
+
+        /// @dev 1. chain id calidation
```
gpersoon, line 58: calidation typo?

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    /// @notice we should only allow the `tradeXChainRFQT` identifier
+    function _decodeTxData(bytes calldata txData_) internal view returns (DecodedQuote memory deBridgeQuote) {
+        InternalVars memory v;
+
+        /// @dev supports both the allowed order types by debridge
+        v.selector = bytes4(txData_[:4]);
+
+        if (v.selector == ICrossChainForwarder.strictlySwapAndCall.selector) {
```
gpersoon, line 173: Function strictlySwapAndCall uses checks on before/after balances but doesn't have a nonreentrancy modifier
There is always a potential risk of abuse in this pattern, although I can't quickly find a real attack
https://vscode.blockscan.com/ethereum/0xe67534a9f24cc000f467eaa17f920bf63b87a2cd

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
```diff
@@ -0,0 +1,154 @@
...
+                !(
+                    receiver == superRegistry.getAddressByChainId(keccak256("CORE_STATE_REGISTRY"), args_.dstChainId)
+                        || hasDstSwap
+                )
+            ) {
+                revert Error.INVALID_TXDATA_RECEIVER();
+            }
+
+            /// @dev if there is a dst swap then the interim token should be the quote of debridge
+            if (hasDstSwap && (args_.liqDataInterimToken != _castToAddress(deBridgeQuote.takeTokenAddress))) {
```
0xTimepunk, line 74: Should we also check the takeTokenAddress is equal to the vault asset address in the case of no dstSwap?

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
0xTimepunk: if we don't intent to allow the user (giver) to increase the amount on source I think we should set givePatchAuthoritySrc to address(0) -> validate with the Debridge team this is correct

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
```diff
@@ -0,0 +1,154 @@
...
+    function validateTxData(ValidateTxDataArgs calldata args_) external view override returns (bool hasDstSwap) {
+        DlnOrderLib.OrderCreation memory deBridgeQuote = _decodeTxData(args_.txData);
+
+        /// sanity check for allowed parameters of tx data
+        if (deBridgeQuote.externalCall.length > 0) revert DeBridgeError.INVALID_EXTRA_CALL_DATA();
+
+        if(deBridgeQuote.allowedTakerDst.length > 0) revert DeBridgeError.INVALID_TAKER_DST();
+
+        if (
+            superRegistry.getAddressByChainId(keccak256("DEBRIDGE_AUTHORITY"), args_.dstChainId)
```
0xTimepunk, line 42: Should agree with @vikramarun and @subhasishgoswami who is the address that is going to be the DEBRIDGE_AUTHORITY

In my opinion this could be the CORE_REGISTRY_UPDATER as it performs similar functionality

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
```diff
@@ -0,0 +1,154 @@
...
+
+    /// @inheritdoc BridgeValidator
+    function validateReceiver(bytes calldata txData_, address receiver) external pure override returns (bool) {
+        DlnOrderLib.OrderCreation memory deBridgeQuote = _decodeTxData(txData_);
+
+        return (receiver == _castToAddress(deBridgeQuote.receiverDst));
+    }
+
+    /// @inheritdoc BridgeValidator
+    /// @dev make sure the OrderCreation.allowedCancelBeneficiarySrc and OrderCreation.givePatchAuthoritySrc is the user address on source
```
0xTimepunk, line 32: I think we should not allow the user to add more input tokens as that can affect the whole intent. Therefore I'd make sure that givePatchAuthoritySrc is address(0)

allowedCancelBeneficiarySrc I think should be validated to be the user address on source (to receive the tokens if any cancelation happens)

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
```diff
@@ -0,0 +1,154 @@
...
+            (deBridgeQuote,,, permitEnvelope) =
+                abi.decode(_parseCallData(txData_), (DlnOrderLib.OrderCreation, bytes, uint32, bytes));
+        } else if (selector == IDlnSource.createSaltedOrder.selector) {
+            (deBridgeQuote,,,, permitEnvelope,) =
+                abi.decode(_parseCallData(txData_), (DlnOrderLib.OrderCreation, uint64, bytes, uint32, bytes, bytes));
+        } else {
+            revert Error.BLACKLISTED_ROUTE_ID();
+        }
+
+        if (permitEnvelope.length > 0) {
```
0xTimepunk, line 140: I assume this is validated to be 0 because tokens are in router and don't need spending ability (are approved in https://github.com/superform-xyz/superform-core/blob/dcc6c1d7b7f4d234b8400a6523ef668bdb977b7b/src/crosschain-liquidity/LiquidityHandler.sol#L50)

src/libraries/DeBridgeError.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.23;
+
+library DeBridgeError {
```
0xTimepunk, line 4: I would move this file to 

crosschain-liquidity/debridge/errors and not the general libraries folder

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,223 @@
...
+    //////////////////////////////////////////////////////////////
+
+    /// @inheritdoc BridgeValidator
+    function validateReceiver(bytes calldata txData_, address receiver) external view override returns (bool) {
+        DecodedQuote memory deBridgeQuote = _decodeTxData(txData_);
+
+        return (receiver == deBridgeQuote.finalReceiver);
+    }
+
+    /// @notice check other parameters including: `givePatchAuthoritySrc`
```
0xTimepunk, line 51: just like in Debridgevalidator.sol comment. Perhaps more things need to be brought back to validateTxData function from the _decodeTxData internal function to achieve this:

> I think we should not allow the user to add more input tokens as that can affect the whole intent. Therefore I'd make sure that givePatchAuthoritySrc is address(0)
> 
> allowedCancelBeneficiarySrc I think should be validated to be the user address on source (to receive the tokens if any cancelation happens)

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
0xTimepunk: Let's only allow cancelation by a keeper IMO

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,223 @@
...
+    /// @inheritdoc BridgeValidator
+    function validateTxData(ValidateTxDataArgs calldata args_) external view override returns (bool hasDstSwap) {
+        DecodedQuote memory deBridgeQuote = _decodeTxData(args_.txData);
+
+        if (
+            superRegistry.getAddressByChainId(keccak256("DEBRIDGE_AUTHORITY"), args_.dstChainId)
```
0xTimepunk, line 57: Just like in debridgevalidator.sol comment 

> Should agree with @vikramarun and @subhasishgoswami who is the address that is going to be the DEBRIDGE_AUTHORITY
> 
> In my opinion this could be the CORE_REGISTRY_UPDATER as it performs similar functionality

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,223 @@
...
+    //////////////////////////////////////////////////////////////
+    //                       CONSTANTS                          //
+    //////////////////////////////////////////////////////////////
+    address private constant DE_BRIDGE_SOURCE = 0xeF4fB24aD0916217251F553c0596F8Edc630EB66;
+
+    //////////////////////////////////////////////////////////////
+    //                       STRUCTS                            //
+    //////////////////////////////////////////////////////////////
+
+    struct DecodedQuote {
```
0xTimepunk, line 24: I guess this was adapted from ICrossChainForwarder

I suggest to rename all inputs to what's present there in the arguments of strictlySwapAndCall and pass the descriptions of the args as comments to here too for clear understanding

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,223 @@
...
+                    deBridgeQuote.finalReceiver
+                        == superRegistry.getAddressByChainId(keccak256("CORE_STATE_REGISTRY"), args_.dstChainId)
+                        || hasDstSwap
+                )
+            ) {
+                revert Error.INVALID_TXDATA_RECEIVER();
+            }
+
+            /// @dev if there is a dst swap then the interim token should be the quote of debridge
+            if (hasDstSwap && (args_.liqDataInterimToken != deBridgeQuote.outputToken)) {
```
0xTimepunk, line 88: Just like in DebridgeValidator.sol

> Should we also check the takeTokenAddress is equal to the vault asset address in the case of no dstSwap?

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,223 @@
...
+
+        /// @dev supports both the allowed order types by debridge
+        v.selector = bytes4(txData_[:4]);
+
+        if (v.selector == ICrossChainForwarder.strictlySwapAndCall.selector) {
+            /// @decode the input txdata
+            (
+                deBridgeQuote.inputToken,
+                deBridgeQuote.inputAmount,
+                ,
```
0xTimepunk, line 154: I think we should validate that _srcTokenInPermitEnvelope is 0

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,223 @@
...
+                ,
```
0xTimepunk, line 155: What is _srcSwapRouter used for? Can it be used against us? doesn't seem to be validated

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,223 @@
...
+                ,
+                v.swapOutputToken,
+                ,
+                v.swapRefundAddress,
```
0xTimepunk, line 159: v.swapRefundAddress is extracted but not validated. I'd send back to the user so I'd validate against it, but check with @vikramarun

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    /// @dev helps parsing debridge calldata and return the input parameters
+    function parseCallData(bytes calldata callData) public pure returns (bytes calldata) {
+        return callData[4:];
+    }
+
+    /// @dev helps parse bytes memory selector
+    function _parseSelectorMem(bytes memory data) internal pure returns (bytes4 selector) {
+        assembly {
```
0xTimepunk, line 214: any reason for using assembly here? This should be consistent with DebridgeValidator.sol

test/mocks/DeBridgeForwarderMock.sol
```diff
@@ -0,0 +1,67 @@
...
+            abi.decode(v.quote.takeTokenAddress, (address)),
+            abi.decode(v.quote.receiverDst, (address)),
+            v.quote.takeAmount
+        );
+
+        vm.selectFork(v.fromChainId);
+    }
+
+    /// @dev helps parse bytes memory selector
+    function _parseCallDataMem(bytes memory data) internal pure returns (bytes memory calldata_) {
```
0xTimepunk, line 62: Should not use assembly if possible

test/unit/crosschain-liquidity/deBridge/DeBridgeValidator.t.sol
```diff
@@ -0,0 +1,394 @@
...
+
+import { Error } from "src/libraries/Error.sol";
+import { DeBridgeError } from "src/libraries/DeBridgeError.sol";
+import "test/utils/ProtocolActions.sol";
+import "src/interfaces/IBridgeValidator.sol";
+import { DeBridgeValidator } from "src/crosschain-liquidity/debridge/DeBridgeValidator.sol";
+import { DlnOrderLib } from "src/vendor/debridge/DlnOrderLib.sol";
+import { IDlnSource } from "src/vendor/debridge/IDlnSource.sol";
+
+contract DeBridgeValidatorTest is ProtocolActions {
```
0xTimepunk, line 12: General comment: could we hardcode a real txData from din api and test it for both DebridgeValidator and Forwarder?

test/utils/TestTypes.sol
```diff
@@ -168,6 +170,9 @@ struct SetupVars {
     address lifiValidator;
     address socketValidator;
     address socketOneInchValidator;
+    address debridgeValidator;
+    address debridgeForwarderValidator;
+    address hashflowValidator;
```
0xTimepunk, line 175: Remove anything in the entire code related to hashflow

0xTimepunk (changes requested):
Some questions to clarify and perhaps extra validations. We're almost there!

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
0xTimepunk: // An optional address on the source (current) chain where the given input tokens
        // would be transferred to in case order cancellation is initiated by the orderAuthorityAddressDst
        // on the destination chain. This property can be safely set to an empty bytes array (0x):
        // in this case, tokens would be transferred to the arbitrary address specified
        // by the orderAuthorityAddressDst upon order cancellation
        bytes allowedCancelBeneficiarySrc;

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
0xTimepunk: You're right, Closing this one

test/mocks/DeBridgeForwarderMock.sol
0xTimepunk: Let's use the same conditions as on source

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
0xTimepunk: I see what you mean, probably not worth the effort. Let's leave it like this

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
0xTimepunk: We could use `receiverAddress` as that is present in ValidateTxDataArgs and make a note that if a user is using a contract wallet that is not create2 this won't work

(because receiverAddress is the address on destination, which for EOAs is the same but for contract accounts may not be)

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
0xTimepunk: To discuss with @vikramarun

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
0xTimepunk: I don't think this is a real risk for us in specific because there is the following validation 
image

Also the user would be able to re-enter only in router functionality for which they would need to put up more tokens to deposit?

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
0xTimepunk: this is srcSender

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
```diff
@@ -0,0 +1,223 @@
...
+                revert Error.INVALID_TXDATA_RECEIVER();
+            }
+
+            /// @dev if there is a dst swap then the interim token should be the quote of debridge
+            if (hasDstSwap && (args_.liqDataInterimToken != deBridgeQuote.outputToken)) {
+                revert Error.INVALID_INTERIM_TOKEN();
+            }
+        } else {
+            /// @dev if withdrawal, then receiver address must be the receiverAddress
+            if (deBridgeQuote.finalReceiver != args_.receiverAddress) revert Error.INVALID_TXDATA_RECEIVER();
```
0xTimepunk, line 93: Warning: receiverAddress is srcSender and this may be a contract that is not create2

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
0xTimepunk: @sujithsomraaj validated allowedCancelBeneficiarySrc to be receiverAddress

wait for @vikramarun on givePatchAuthoritySrc

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
subhasishgoswami: Lets make it CORE_STATE_REGISTRY_RESCUER_ROLE, will be in line with whole rescue process

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
gpersoon: the routers are also whitelisted so there should be enough barriers

src/crosschain-liquidity/debridge/DeBridgeForwarderValidator.sol
0xTimepunk: Ask debridge to see if they have a registry of swapRouters, otherwise hard to maintain/validate

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
vikramarun: we could support this the same way we will have makePayment in activity page. 
but it should be the src sender user address

src/crosschain-liquidity/debridge/DeBridgeValidator.sol
vikramarun: we can support this the same way we would makePayment in activity page. let's include but have it be the src sender

script/forge-scripts/EnvironmentUtils.s.sol
```diff
@@ -35,6 +35,7 @@ abstract contract EnvironmentUtils is AbstractDeploySingle {
             TARGET_CHAINS.push(ARBI);
             TARGET_CHAINS.push(OP);
             TARGET_CHAINS.push(BASE);
+            TARGET_CHAINS.push(FANTOM);
```
0xTimepunk, line 38: Remove fantom from here as it is not deployed into production

script/forge-scripts/misc/Abstract.Deploy.DeBridgeValidators.s.sol
```diff
@@ -0,0 +1,109 @@
...
+        newBridgeids[1] = 8;
+
+        assert(NEW_BRIDGE_ADDRESSES[vars.chainId][0] != address(0));
+        assert(NEW_BRIDGE_ADDRESSES[vars.chainId][1] != address(0));
+        assert(bridgeValidators[0] != address(0));
+        assert(bridgeValidators[1] != address(0));
+
+        vars.superRegistryC =
+            SuperRegistry(payable(_readContractsV1(env, chainNames[trueIndex], vars.chainId, "SuperRegistry")));
+        address expectedSr = env == 0
```
0xTimepunk, line 100: If this function is only going to be called by staging we don't need such if else

can be just 

```
address expectedSr = vars.chainId == 250 ? 0x7B8d68f90dAaC67C577936d3Ce451801864EF189 : 0xB2C097ac459aFAc892ae5b35f6bd6a9Dd3071F47;
```

script/utils/misc/run_script_mainnet_deBridge.sh
```diff
@@ -0,0 +1,39 @@
...
+wait
+
+FOUNDRY_PROFILE=production forge script script/forge-scripts/misc/Mainnnet.DeployDeBridgeValidators.s.sol:MainnetDeployDeBridgeValidators --sig "deployDeBridgeValidator(uint256,uint256)" 0 5 --rpc-url $OPTIMISM_RPC_URL --slow --account defaultKey --sender 0x48aB8AdF869Ba9902Ad483FB1Ca2eFDAb6eabe92 --legacy
+wait
+
+FOUNDRY_PROFILE=production forge script script/forge-scripts/misc/Mainnnet.DeployDeBridgeValidators.s.sol:MainnetDeployDeBridgeValidators --sig "deployDeBridgeValidator(uint256,uint256)" 0 6 --rpc-url $BASE_RPC_URL --slow --account defaultKey --sender 0x48aB8AdF869Ba9902Ad483FB1Ca2eFDAb6eabe92 --legacy
+wait
+
+FOUNDRY_PROFILE=production forge script script/forge-scripts/misc/Mainnnet.DeployDeBridgeValidators.s.sol:MainnetDeployDeBridgeValidators --sig "deployDeBridgeValidator(uint256,uint256)" 0 7 --rpc-url $FANTOM_RPC_URL --slow --account defaultKey --sender 0x48aB8AdF869Ba9902Ad483FB1Ca2eFDAb6eabe92 --legacy
```
0xTimepunk, line 37: Remove this fantom line, for now. Fantom is not on production

script/forge-scripts/Abstract.Deploy.Single.s.sol
```diff
@@ -100,7 +104,7 @@ abstract contract AbstractDeploySingle is BatchScript {
     address public constant CANONICAL_PERMIT2 = 0x000000000022D473030F116dDEE9F6B43aC78BA3;
     mapping(uint64 chainId => mapping(bytes32 implementation => address at)) public contracts;
 
-    string[22] public contractNames = [
+    string[24] public contractNames = [
```
0xTimepunk, line 107: This file (Abstract.Deploy.Single.s.sol) needs to have the deployments of DebridgeValidator and debridge forwarderValidator (as it will be used when deploying fantom to production from here, for e.g)

0xTimepunk (changes requested):
A few more things to cleanup

0xTimepunk (approved):
LGTM

PR #503, opened 2024-04-12
feat: SUP-6433 Testing Rewards Distributor

0xTimepunk:
This PR contains
- Invariant test for merkle tree full claim
- Invariant test for batch merkle tree full claim
- Full coverage
- Refactoring of invariant folder
- Fix of unknown bug in vault shares invariant test (opens possibility of further debugging of other bugs)


src/RewardsDistributor.sol
```diff
@@ -104,7 +104,7 @@ contract RewardsDistributor is IRewardsDistributor {
         uint256 tokensToClaim = rewardTokens_.length;
 
         if (tokensToClaim == 0) revert ZERO_ARR_LENGTH();
-        if (tokensToClaim != amountsClaimed_.length) revert INVALID_BATCH_REQ();
+        if (tokensToClaim != amountsClaimed_.length) revert INVALID_REQ_TOKENS_AMOUNTS();
```
gpersoon, line 107: could also check `proof_.length`

src/RewardsDistributor.sol
```diff
@@ -162,11 +164,12 @@ contract RewardsDistributor is IRewardsDistributor {
         if (root == ZERO_BYTES32) revert MERKLE_ROOT_NOT_SET();
 
         /// @dev user cannot claim a periodic reward twice
-        if (periodicRewardsClaimed[periodId_][claimer_]) return false;
+        if (periodicRewardsClaimed[periodId_][claimer_]) revert ALREADY_CLAIMED();
 
         bytes32 leaf = keccak256(
             bytes.concat(keccak256(abi.encode(claimer_, periodId_, rewardTokens_, amountsClaimed_, CHAIN_ID)))
         );
+
```
gpersoon, line 172: To save some gas could use [OZ verifyCalldata()](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/MerkleProof.sol#L41)

src/RewardsDistributor.sol
gpersoon, line 107: What if the receiver can't (currently) receive the tokens; because it is a smart contract,
or it is a smart contract (gnosis safe), but deployed on another chain

The permissionless claim for random receivers might make that worse

gpersoon, line 107: `setPeriodicRewards()`:
Mistakes in merkle tree can't be fixed
Might be relevent if the receiver can't receive the tokens (private key lost, smart contract)

(good for permissionlessness, but the RewardsAdmin is a trusted entity anyway)

gpersoon, line 107: `claim()`
The claims can be done for ever (no deadline)
Good for permissionlessness, but perhaps you want to redistribute unclaimed tokens

src/RewardsDistributor.sol
gpersoon, line 172: Could leave a comment on why the double hashing is used, for example:
https://www.rareskills.io/post/merkle-tree-second-preimage-attack

src/RewardsDistributor.sol
0xTimepunk: If there are any mistakes in the tree and tokens could not be claimed this could be fixed with a new corrective merkle tree

src/RewardsDistributor.sol
0xTimepunk: Added

src/RewardsDistributor.sol
0xTimepunk: >What if the receiver can't (currently) receive the tokens; because it is a smart contract

We thought of changing msg.sender in https://github.com/superform-xyz/superform-core/blob/434915c83424277fe0d1c66d5570c332491f93a9/src/RewardsDistributor.sol#L191 to `receiver_` such that anyone can submit proofs on behalf of a receiver_  and tokens would be sent to that receiver_

This would allow an EOA to claim tokens on behalf of the contract, avoiding the contract needing to have the claim function implementation

Is this related to "The permissionless claim for random receivers might make that worse" ?

What do you think of this change?

> or it is a smart contract (gnosis safe), but deployed on another chain

We don't intend to support smart contract wallets that have different addresses on different chains. There will be an acknowledgement by the user that we do not provide reward claiming on this situation.

gpersoon: I thought `verifyClaim()` is called with `receiver_`, hence my remark.
However checking the code again this isn't the case and I think there is a bigger issue here:

Suppose `Claim()` is called for `receiver1`:
`verifyClaim()` checks `periodicRewardsClaimed[][msg.sender]`, which is false
and then sends the tokens to `receiver1`  and sets `periodicRewardsClaimed[][receiver1]` to true.

Suppose then the same claimer calls `Claim()` is called for `receiver2`:
`verifyClaim()` checks `periodicRewardsClaimed[][msg.sender]` , which is false!!!!!!
and then sends the tokens to `receiver2`  and sets `periodicRewardsClaimed[][receiver2]` to true.

```diff
function _claim(...) ... {
    if (!verifyClaim(msg.sender, periodId_, rewardTokens_, amountsClaimed_, proof_)) revert INVALID_CLAIM();
-   periodicRewardsClaimed[periodId_][receiver_] = true;
+   periodicRewardsClaimed[periodId_][msg.sender] = true;
    _transferRewards(receiver_, rewardTokens_, amountsClaimed_, tokensToClaim_);
}
```

0xTimepunk: I think there might be a misunderstanding here so allow me to explain better


The check
`        if (periodicRewardsClaimed[periodId_][claimer_]) revert ALREADY_CLAIMED();`

is on **claimer_**, not on msg.sender. 

Therefore, in your examples and 

Suppose a msg.sender calls Claim() for receiver1:
claimer_ = receiver1
- verifyClaim() checks periodicRewardsClaimed[][claimer_], which is false
- and then sends the tokens to receiver1 and sets periodicRewardsClaimed[][receiver1] to true.

Suppose then the same msg.sender calls Claim() with receiver2`:
claimer_ = receiver2

- verifyClaim() checks periodicRewardsClaimed[][claimer_] . 
- If at this point receiver2 had a valid claims then tokens would be sent o receiver2 and periodicRewardsClaimed[][receiver2] to true.


I think the misunderstanding came from the usage of the word "claimer". Msg.sender can be any

gpersoon: Oh wait, I didn't see that you already had proposed changing `msg.sender` to `receiver_`. That should indeed work.
The address that is used for the merke tree check 
should also be the address that is used in `periodicRewardsClaimed[][]`

To summarize the code should be:
```diff
function _claim(...) ... {
-   if (!verifyClaim(msg.sender, periodId_, rewardTokens_, amountsClaimed_, proof_)) revert INVALID_CLAIM();
+   if (!verifyClaim(receiver_, periodId_, rewardTokens_, amountsClaimed_, proof_)) revert INVALID_CLAIM();
    periodicRewardsClaimed[periodId_][receiver_] = true;
    _transferRewards(receiver_, rewardTokens_, amountsClaimed_, tokensToClaim_);
}
```
```diff
-function verifyClaim(address claimer_, ... ) ... {
    ...
-   if (periodicRewardsClaimed[periodId_][claimer_]) revert ALREADY_CLAIMED();
-   bytes32 leaf = keccak256(bytes.concat(keccak256(abi.encode(claimer_, ...))) );
+function verifyClaim(address receiver_, ... ) ... {
    ...
+   if (periodicRewardsClaimed[periodId_][receiver_]) revert ALREADY_CLAIMED();
+   bytes32 leaf = keccak256(bytes.concat(keccak256(abi.encode(receiver_, ...))) );
   ...
}
```

A potential risk would be that rewards are claimed for someone who maybe doesn't want the rewards to be claimed already. Might be due to tax reasons, or maybe the receiving address can't remove the tokens.

PR #505, opened 2024-04-19
feat: Optimize payment helper SUP-6550

0xTimepunk:
Reduces size under the maximum limit to deploy on mainnet and makes code much more maintainable.

Fixes a bug that was present in in estimateMultiDstMultiVault
where we are adding to srcAmount

```
srcAmount += _estimateAckProcessingCost(v.superformIdsLen);
```

but should be instead: 

```
srcAmount += _estimateAckProcessingCost(ackLen);
```

### DIFF
--- a/src/payments/PaymentHelper.sol
+++ b/src/payments/PaymentHelper.sol
@@ -102,6 +102,16 @@ contract PaymentHelper is IPaymentHelper {
         bool paused;
     }
 
+    struct CalculateAmountsReq {
+        uint256 i;
+        uint64[] dstChainIds;
+        uint8[] ambIds;
+        MultiVaultSFData[] superformsData;
+        SingleVaultSFData[] superformData;
+        ISuperformFactory factory;
+        bool isDeposit;
+    }
+
     //////////////////////////////////////////////////////////////
     //                       MODIFIERS                          //
     //////////////////////////////////////////////////////////////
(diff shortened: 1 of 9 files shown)


src/payments/PaymentHelper.sol
```diff
@@ -1175,4 +1020,126 @@ contract PaymentHelper is IPaymentHelper {
...
+                if (multiVaults) {
+                    for (uint256 j; j < v.superformIdsLen; ++j) {
+                        if (!req_.superformsData[req_.i].retain4626s[j]) ++ackLen;
+                    }
+                } else {
+                    if (!req_.superformData[req_.i].retain4626) ++ackLen;
+                }
+
+                /// @dev step 4: estimation processing cost of acknowledgement on source
+                srcAmount += _estimateAckProcessingCost(ackLen);
```
gpersoon, line 1068: The previous version of `estimateMultiDstMultiVault()` was different, but I guess that was a bug.
```solidity
function estimateMultiDstMultiVault(
    ...
    for (uint256 j; j < v.superformIdsLen; ++j) {
        if (!req_.superformsData[i].retain4626s[j]) ++ackLen;
    }
    srcAmount += _estimateAckProcessingCost(v.superformIdsLen); // doesn't use ackLen
    ...
}
```

src/payments/PaymentHelper.sol
```diff
@@ -1175,4 +1020,126 @@ contract PaymentHelper is IPaymentHelper {
...
+
+        if (!paused && formId == TIMELOCK_FORM_ID) {
+            totalDstGas += timelockCost[dstChainId_];
+        } else if (paused) {
+            totalDstGas += emergencyCost[dstChainId_];
+        }
+    }
+
+    /// @dev calculates the srcAmount cost for single direct withdrawal
+    function _calculateSrcAmountTimelockEmergencyDirect(
```
gpersoon, line 1124: _calculateSrcAmountTimelockEmergencyDirect and _calculateTotalDstGasTimelockEmergency
are similar, maybe the can be integrated too, if you need to save more contract space

src/payments/PaymentHelper.sol
```diff
@@ -102,6 +102,16 @@ contract PaymentHelper is IPaymentHelper {
         bool paused;
     }
 
+    struct CalculateAmountsReq {
+        uint256 i;
+        uint64[] dstChainIds;
+        uint8[] ambIds;
+        MultiVaultSFData[] superformsData;
+        SingleVaultSFData[] superformData;
```
gpersoon, line 110: I think `SingleVaultSFData` doesn't have to be an array. Not sure it the would save
deployment or execution gas though because then the entire struct `SingleVaultSFData` is included

src/payments/PaymentHelper.sol
0xTimepunk: Yes, it was a bug, this should be with ackLen, mentioned in original message of this PR!

src/payments/PaymentHelper.sol
0xTimepunk: That's what I thought hence why I did it this way - think I am going to keep it like this

src/payments/PaymentHelper.sol
```diff
@@ -923,6 +931,9 @@ contract PaymentHelper is IPaymentHelper {
         if (dstNativeFee == 0) {
             return 0;
         }
+        if (!xChain_) {
+            return dstNativeFee;
+        }
```
gpersoon, line 936: I think on line 950 `nativeTokenPrice` could be used instead of `_getNativeTokenPrice(CHAIN_ID)`

src/payments/PaymentHelper.sol
```diff
@@ -914,7 +914,15 @@ contract PaymentHelper is IPaymentHelper {
     /// @dev helps convert the dst gas fee into src chain native fee
     /// @dev https://docs.soliditylang.org/en/v0.8.4/units-and-global-variables.html#ether-units
     /// @dev all native tokens should be 18 decimals across all EVMs
-    function _convertToNativeFee(uint64 dstChainId_, uint256 dstGas_) internal view returns (uint256 nativeFee) {
+    function _convertToNativeFee(
```
gpersoon, line 917: `_convertToSrcNativeAmount()` is similar to `_convertToNativeFee()` is further contract space is useful maybe these can be combined too

0xTimepunk: We think that, despite similar, these functions perform a bit different functionality. _convertToSrcNativeAmount is called from a destination chain and doesn't include a getGasPrice call - the cost (dstAmount_) is Brought in by estimateAckCostDefault


At least for now we feel this should remain as is

PR #516, opened 2024-05-02
feat: Draft erc7540

0xTimepunk:


src/crosschain-data/extensions/AsyncStateRegistry.sol
```diff
@@ -0,0 +1,397 @@
...
+        if (data_.receiverAddress == address(0)) revert Error.RECEIVER_ADDRESS_NOT_SET();
+
+        ++asyncWithdrawPayloadCounter;
+
+        asyncWithdrawPayload[asyncWithdrawPayloadCounter] =
+            AsyncWithdrawPayload(type_, srcChainId_, requestId_, data_, AsyncStatus.PENDING);
+    }
+
+    /// @inheritdoc IAsyncStateRegistry
+    function finalizeDepositPayload(uint256 asyncPayloadId_) external payable override onlyAsyncStateRegistry {
```
vikramarun, line 173: if im understanding this right, there would just be one keeper right? could we called it AsyncStateRegistryProcessor? (this can finalize payloads)

src/crosschain-data/extensions/AsyncStateRegistry.sol
```diff
@@ -0,0 +1,397 @@
...
+        /// @dev restoring state for gas saving
+        /// @dev should this be done ? considering we'd want to mark this as forever not possible to process in case of
+        /// failure?
+        /// @dev or maybe the form is momentarily paused and we'd want the keeper to try again later when it's
+        /// unpaused...
+        delete asyncWithdrawPayload[asyncPayloadId_];
+    }
+
+    /// @inheritdoc BaseStateRegistry
+    function processPayload(uint256 payloadId_)
```
vikramarun, line 309: this is called before `finalize*`?

src/crosschain-data/extensions/AsyncStateRegistry.sol
```diff
@@ -0,0 +1,397 @@
...
+                }
+
+                /// @dev for direct chain, superPositions are minted directly
+                if (p.isXChain == 0) {
+                    ISuperPositions(superRegistry.getAddress(keccak256("SUPER_POSITIONS"))).mintSingle(
+                        p.data.receiverAddress, p.data.superformId, shares
+                    );
+                }
+            }
+        } catch {
```
vikramarun, line 214: so pretty much if we try to finalize but it fails, none of the state changes, and we have to try again? (this makes sense)

src/crosschain-data/extensions/AsyncStateRegistry.sol
```diff
@@ -0,0 +1,397 @@
...
+                revert Error.SLIPPAGE_OUT_OF_BOUNDS();
+            }
+
+            p.data.liqData.txData = txData_;
+        }
+
+        /// @dev if redeeming failed superPositions are not reminted
+        /// @dev this is different than the normal 4626 flow because if a redeem is claimable
+        /// @dev a user could simply go to the vault and claim the assets directly
+        superform.claimWithdraw(p);
```
vikramarun, line 298: no try/catch here? if the redemption fails we would have to try again (but no event is emitted?)

src/forms/ERC7540Form.sol
```diff
@@ -0,0 +1,547 @@
...
+        } else {
+            /// @dev transfer shares to user and do not redeem shares for assets
+            IERC20(IERC7540(vault).share()).safeTransfer(singleVaultData_.receiverAddress, singleVaultData_.amount);
+        }
+
+        return 0;
+    }
+
+    /// @inheritdoc BaseForm
+    function _emergencyWithdraw(address, uint256) internal virtual override {
```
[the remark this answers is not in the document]
0xTimepunk: Makes sense this case, for when the vault share is held and the user initiates the first step of the withdraw (the request step)

src/crosschain-data/extensions/AsyncStateRegistry.sol
0xTimepunk: Can finalize and process payloads

src/crosschain-data/extensions/AsyncStateRegistry.sol
0xTimepunk: Nope.

After the amounts are claimable, someone (a keeper) has to call the claim function. This is done via the `finalize` step.

If a claimDeposit is successful (shares minted for superform), a message callback will be issue back to the source chain.

Then the same AsyncStateRegistryProcessor has to call the processPayload function to mint superPositions.

src/crosschain-data/extensions/AsyncStateRegistry.sol
0xTimepunk: @sujithsomraaj in the case funds are still in a request phase in the vault (not claimable yet), an option would be to add the cancelation flow, which is supported by a separate EIP. But @vikramarun said not to think about this right now (but perhaps we should add)

src/crosschain-data/extensions/AsyncStateRegistry.sol
0xTimepunk: It is very risky to enable try catch here.

Imagine the following situation:
-  claimWithdraw returns 0 - this would revert at the form level in https://github.com/superform-xyz/superform-core/blob/bb1e68808b40271f3a85e7802c0401b5acbc84f3/src/forms/ERC7540Form.sol#L215
- If we had try catch and remint, it would remint superPositions for the user.
- Then, imagine suddenly assets starts to return non 0 - the user could still go directly to the vault and claim the withdraw (because maybe it doesn't revert by calling it there) and still have superPositions, breaking the invariant and stealing funds


@vikramarun if this function reverts, backend could try again after some time. However the rationale is that claimWithdraw must pass flawlessly otherwise there's a risk for us

src/forms/ERC7540Form.sol
```diff
@@ -0,0 +1,547 @@
...
+    //////////////////////////////////////////////////////////////
+    //                         ERRORS                         //
+    //////////////////////////////////////////////////////////////
+    error NOT_ASYNC_STATE_REGISTRY();
+
+    //////////////////////////////////////////////////////////////
+    //                         CONSTANTS                         //
+    //////////////////////////////////////////////////////////////
+
+    uint8 constant stateRegistryId = 2; // AsyncStateRegistry
```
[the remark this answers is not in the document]
0xTimepunk: Since we don't have it private in other forms I'd leave it as is for consistency

src/forms/ERC7540Form.sol
```diff
@@ -0,0 +1,547 @@
...
+        uint256 requestId,
+        address owner
+    )
+        public
+        view
+        virtual
+        override
+        returns (uint256 pendingAssets)
+    {
+        return IERC7540(vault).pendingDepositRequest(requestId, owner);
```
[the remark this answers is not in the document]
0xTimepunk: Nice catch. Wrote about this to the alliance guys

src/forms/ERC7540Form.sol
```diff
@@ -0,0 +1,547 @@
...
+
+    /// @inheritdoc IERC7540FormBase
+    function claimDeposit(AsyncDepositPayload memory p_) external onlyAsyncStateRegistry returns (uint256 shares) {
+        if (p_.data.receiverAddress == address(0)) revert Error.RECEIVER_ADDRESS_NOT_SET();
+
+        if (_isPaused(p_.data.superformId)) {
+            /// @dev in case of a deposit and the form is paused, nothing can be sent to the emergency queue as there
+            /// are no shares belonging to this
+            /// @dev payload in the superform. return 0 to stop processing
+            return 0;
```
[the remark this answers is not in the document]
0xTimepunk: In the claim step there are no shares in the superform belonging to this user and assets have already been deposited.

If he wanted, he can skip superform and claim the deposit directly in the vault, bypassing anything here including emergency queue. 

I think we should not have emergency queue processing in claim functions, nor refund flow.

src/forms/ERC7540Form.sol
```diff
@@ -0,0 +1,547 @@
...
+        uint256 requestId,
+        address owner
+    )
+        public
+        view
+        virtual
+        override
+        returns (uint256 claimableAssets)
+    {
+        return IERC7540(vault).claimableDepositRequest(requestId, owner);
```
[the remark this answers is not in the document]
0xTimepunk: Nice catch. Wrote about this to the alliance guys and they will fix it

src/crosschain-data/extensions/AsyncStateRegistry.sol
```diff
@@ -0,0 +1,397 @@
...
+                /// @dev for direct chain, superPositions are minted directly
+                if (p.isXChain == 0) {
+                    ISuperPositions(superRegistry.getAddress(keccak256("SUPER_POSITIONS"))).mintSingle(
+                        p.data.receiverAddress, p.data.superformId, shares
+                    );
+                }
+            }
+        } catch {
+            /// @dev In case of a deposit actual failure (at the vault level, or returned shares level in the form),
+            /// @dev the course of action for a user to claim the deposit would be to directly call claim deposit at the
```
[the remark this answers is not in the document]
0xTimepunk: Yes they can because the user is the requester, aka the owner of the request. So they have the power to go directly to the vault and call claim when it's ready if they wanted (they don't need superform for that)

src/forms/ERC7540Form.sol
```diff
@@ -0,0 +1,547 @@
...
+            revert Error.VAULT_IMPLEMENTATION_FAILED();
+        }
+    }
+
+    /// @inheritdoc IERC7540FormBase
+    function claimWithdraw(AsyncWithdrawPayload memory p_) external onlyAsyncStateRegistry returns (uint256 assets) {
+        if (p_.data.receiverAddress == address(0)) revert Error.RECEIVER_ADDRESS_NOT_SET();
+
+        if (_isPaused(p_.data.superformId)) {
+            /// @dev in case of a deposit and the form is paused, nothing can be sent to the emergency queue as there
```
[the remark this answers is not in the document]
0xTimepunk: cleared text

src/crosschain-data/extensions/AsyncStateRegistry.sol
0xTimepunk: Scenario:
1- user frontruns directly the claim
2- claimable amount drops to 0
3- how does superform stop processing this payload since it was directly claimed, but no FailedDeposit event was emitted

test/mocks/ERC7540Mock.sol
```diff
@@ -0,0 +1,376 @@
...
+        validateController(controller);
+        require(assets == assetBalances[controller][requestId][1], "ERC7540Vault/invalid-deposit-claim");
+
+        assetBalances[controller][requestId][1] = 0;
+
+        shares = convertToShares(assets);
+        ERC7575Mock(share).mint(receiver, shares);
+    }
+
+    /// @inheritdoc IERC7575
```
vikramarun, line 283: should this be from 7575 or 4626? probably fine either way but i guess 7540 should have mint/withdraw implemented at all

test/mocks/ERC7540Mock.sol
```diff
@@ -0,0 +1,376 @@
...
+    function deposit(uint256 assets, address receiver) external returns (uint256 shares) {
+        shares = deposit(assets, receiver, msg.sender);
+    }
+
+    /// @inheritdoc IERC7575
+    function maxMint(address) public pure returns (uint256 maxShares) {
```
vikramarun, line 289: should this be implemented?

test/mocks/ERC7540Mock.sol
```diff
@@ -0,0 +1,376 @@
...
+    uint256 public immutable deploymentChainId;
+    bytes32 private immutable _DOMAIN_SEPARATOR;
+    bytes32 public constant AUTHORIZE_OPERATOR_TYPEHASH = keccak256(
+        "AuthorizeOperator(address controller,address operator,bool approved,uint256 validAfter,uint256 validBefore,bytes32 nonce)"
+    );
+
+    mapping(address controller => mapping(bytes32 nonce => bool used)) authorizations;
+
+    /// @dev state 0 is for pending, state 1 is claimable, state 2 is claimed
+    mapping(address controller => mapping(uint256 requestId => mapping(uint8 state => uint256 assetBalance)))
```
vikramarun, line 44: state 0 for pending, 1 for claimable, etc -- is this how it is in Centifuge? i guess there is no standardization around how this mapping is being done in the spec

0xTimepunk: This is really a mock I came up with to control movement of balances. It is not intended to reflect Centrifuge's operating mechanism nor anything in the spec

Basically the tests will call either moveSharesToClaimable or moveAssetsToClaimable to move from request step to claimable step

test/mocks/ERC7540Mock.sol
0xTimepunk: Nope, not needed

test/mocks/ERC7540Mock.sol
0xTimepunk: Didn't import from IERC4626 because it would force me to implement other functions. Using from 7575 instead

src/forms/ERC7540Form.sol
```diff
@@ -0,0 +1,684 @@
...
+        if (deposit && asset != NATIVE) {
+            _assetTransferIn(asset, amountIn);
+        }
+
+        _validateTxData(bridgeValidator, args);
+
+        _dispatchTokens(bridgeValidator, txData, asset, amountIn, nativeAmount);
+    }
+
+    function _claim(
```
gpersoon, line 664: the name `_claim` is a bit confusing. For redeem it is ok, but for deposit is doesn't seem appropriate

src/forms/ERC7540Form.sol
```diff
@@ -0,0 +1,684 @@
...
+    {
+        return IERC7540(vault).claimableRedeemRequest(requestId, owner);
+    }
+
+    //////////////////////////////////////////////////////////////
+    //              EXTERNAL WRITE FUNCTIONS                    //
+    //////////////////////////////////////////////////////////////
+
+    /// @inheritdoc IERC7540FormBase
+    function claimDeposit(AsyncDepositPayload memory p_) external onlyAsyncStateRegistry returns (uint256 shares) {
```
gpersoon, line 169: doublecheck: should _vaultKind be checked here? (same in claimWithdraw)    
(not sure if it is possible to end up here with the wrong type)

src/forms/ERC7540Form.sol
```diff
@@ -0,0 +1,684 @@
...
+
+        IERC7540 v = IERC7540(vault);
+
+        address sharesReceiver = p_.data.retain4626 ? p_.data.receiverAddress : address(this);
+
+        uint256 sharesBalanceBefore;
+        uint256 sharesBalanceAfter;
+
+        (sharesBalanceBefore, sharesBalanceAfter, shares) =
+            _claim(v, v.share(), p_.data.amount, sharesReceiver, p_.data.receiverAddress, true);
```
gpersoon, line 188: The supplied controller is `p_.data.receiverAddress`
When looking at  redeem / deposit
https://github.com/centrifuge/liquidity-pools/blob/main/src/ERC7540Vault.sol#L348
https://github.com/centrifuge/liquidity-pools/blob/main/src/ERC7540Vault.sol#L300

it calls validateController
https://github.com/centrifuge/liquidity-pools/blob/main/src/ERC7540Vault.sol#L401

which checks it is msg.sender or an authorized address.
So I'm not sure the current implementation would work  (should show in the tests)
Same for claimWithdraw()

src/forms/ERC7540Form.sol
```diff
@@ -0,0 +1,684 @@
...
+    )
+        internal
+    {
+        address bridgeValidator = _getBridgeValidator(bridgeId);
+
+        uint256 amountIn = _decodeAmountIn(bridgeValidator, txData);
+
+        if (deposit && asset != NATIVE) {
+            _assetTransferIn(asset, amountIn);
+        }
```
gpersoon, line 657: in case of claimWithdraw, should _swapAssetsInOrOut do _assetTransferIn?
Because withdrawAfterCoolDown didn't do that

(solution would be to remove the call to _assetTransferIn from swapAssetsInOrOut
and in _requestDirectDeposit do _assetTransferIn indifferent of singleVaultData_.liqData.txData.length )

src/forms/ERC7540Form.sol
```diff
@@ -0,0 +1,684 @@
...
+    }
+
+    function _vaultKindCheck() public view returns (VaultKind kind) {
+        bool depositSupported;
+        bool redeemSupported;
+
+        /// @dev ideally the check is made at the selector level
+        try IERC165(vault).supportsInterface(type(IERC7540Deposit).interfaceId) returns (bool depositSupported_) {
+            depositSupported = depositSupported_;
+            try IERC165(vault).supportsInterface(type(IERC7540Redeem).interfaceId) returns (bool redeemSupported_) {
```
gpersoon, line 579: this try doesn't have to be nested because the previous try will revert if it fails

src/forms/ERC7540Form.sol
```diff
@@ -0,0 +1,684 @@
...
+            } catch {
+                revert ERC_165_INTERFACE_REDEEM_CALL_FAILED();
+            }
+        } catch {
+            revert ERC_165_INTERFACE_DEPOSIT_CALL_FAILED();
+        }
+
+        if (depositSupported && redeemSupported) {
+            return VaultKind.FULLY_ASYNC;
+        } else if (depositSupported) {
```
gpersoon, line 590: the `else`s are not necessary because a return is done

src/forms/ERC7540Form.sol
```diff
@@ -0,0 +1,684 @@
...
+
+        assetsToDeposit = IERC20(vars.asset).balanceOf(address(this)) - vars.balanceBefore;
+
+        /// @dev the difference in vault tokens, ready to be deposited, is compared with the amount inscribed in the
+        /// superform data
+        if (
+            assetsToDeposit * ENTIRE_SLIPPAGE
+                < singleVaultData_.amount * (ENTIRE_SLIPPAGE - singleVaultData_.maxSlippage)
+        ) {
+            revert Error.DIRECT_DEPOSIT_SWAP_FAILED();
```
gpersoon, line 475: could maybe use `_slippageValidation()`

src/crosschain-data/extensions/AsyncStateRegistry.sol
```diff
@@ -0,0 +1,402 @@
...
+        }
+
+        (address superformAddress,,) = p.data.superformId.getSuperform();
+
+        IERC7540Form superform = IERC7540Form(superformAddress);
+
+        uint256 claimableDeposit = superform.getClaimableDepositRequest(p.requestId, p.data.receiverAddress);
+        if (
+            p.requestId == 0 && claimableDeposit < p.assetsToDeposit
+                || p.requestId != 0 && claimableDeposit != p.assetsToDeposit
```
gpersoon, line 191: safe to put some `(` `)`   to prevent evaluation errors (`&&` versus `||`)

src/crosschain-data/extensions/AsyncStateRegistry.sol
```diff
@@ -0,0 +1,402 @@
...
+                    );
+                }
+
+                /// @dev for direct chain, superPositions are minted directly
+                if (p.isXChain == 0) {
+                    ISuperPositions(superRegistry.getAddress(keccak256("SUPER_POSITIONS"))).mintSingle(
+                        p.data.receiverAddress, p.data.superformId, shares
+                    );
+                }
+            }
```
gpersoon, line 218: if shares == 0 shouldn't there be some error handling? 
p.status is already set to AsyncStatus.PROCESSED so can't be done again

src/crosschain-data/extensions/AsyncStateRegistry.sol
```diff
@@ -0,0 +1,402 @@
...
+                || p.requestId != 0 && claimableDeposit != p.assetsToDeposit
+        ) {
+            revert NOT_READY_TO_CLAIM();
+        }
+
+        /// @dev set status here to prevent re-entrancy
+        p.status = AsyncStatus.PROCESSED;
+
+        try superform.claimDeposit(p) returns (uint256 shares) {
+            if (shares != 0 && !p.data.retain4626) {
```
gpersoon, line 200: doublecheck: no slippage check?

src/crosschain-data/extensions/AsyncStateRegistry.sol
```diff
@@ -0,0 +1,402 @@
...
+        AMBMessage memory _message = AMBMessage(_payloadHeader, _payloadBody);
+
+        /// @dev validates quorum
+        bytes32 _proof = _message.computeProof();
+
+        if (messageQuorum[_proof] < _getRequiredMessagingQuorum(srcChainId)) {
+            revert Error.INSUFFICIENT_QUORUM();
+        }
+
+        if (callbackType == uint256(CallbackType.RETURN)) {
```
gpersoon, line 342: doublecheck, in timelockstateregistry this was FAIL

same in `_constructSingleReturnData`

src/forms/ERC7540Form.sol
0xTimepunk: In 7540, the claim action is attributed both to deposits or redeems: It means the claiming of your fullfilled deposit/redeem request, hence the name choice

src/forms/ERC7540Form.sol
0xTimepunk: Such check can't hurt, will add

src/forms/ERC7540Form.sol
0xTimepunk: The goal is to authorise this Form (msg.sender) to be an operator on behalf of controller `p_.data.receiverAddress`

This will be done by using this function from this interface https://github.com/centrifuge/liquidity-pools/blob/33a142ca3861bbbc1e3c385afc9002f4ea150e10/src/interfaces/IERC7540.sol#L241-L248

src/forms/ERC7540Form.sol
0xTimepunk: In claimWithdraw we pass the flag "false" to the `bool deposit` of `_swapAssetsInOrOut`, therefore _assetTransferIn is not called.

src/forms/ERC7540Form.sol
0xTimepunk: I thought about this, but only the 2nd parcel of what's done in `_slippageValidation()` is done in L472-473

src/crosschain-data/extensions/AsyncStateRegistry.sol
0xTimepunk: If shares == 0, the goal is not to mint any superPositions. Perhaps an event can be edited in this case, to confirm the misbehaviour? Didn't want to revert the function for this case to leave the state change in p.status

src/crosschain-data/extensions/AsyncStateRegistry.sol
0xTimepunk: I would say if the check is done in claimDeposit with `_slippageValidation` then it is not necessary here as it wastes gas, what do you think?

src/crosschain-data/extensions/AsyncStateRegistry.sol
0xTimepunk: This is because this AsyncStateRegistry operates under a different assumption than the TimelockStateRegistry. 

In AsyncStateRegistry it is assumed that, if a deposit is successful, an acknolwedgement is emitted back to the source chain and a payload must be processed for SuperPositions to be minted. If shares == 0 in a deposit nothing is done.

However, in terms of withdraws (which was the only flow Supported in TimelockStateRegistry), now it is assumed that if a withdraw fails at the claimWithdraw state, then `finalizeWithdrawPayload`will simply revert (there is no try catch mechanism there). SuperPositions would never be reminded. This is a security mechanism to prevent an issue where a claimWithdraw is temporary unavailable due to some reason. Later, if it becomes unavailable again a user (who is a controller of the request), could go to the vault directly and claim withdraw, and if there was a try catch together with an acknowledgement of CallbackType.FAIL, it could remind SuperPositions back to the user in case of a new failure, even though the user already claimed assets directly

src/forms/ERC7540Form.sol
gpersoon: oh yes this indeed takes care of this

src/crosschain-data/extensions/AsyncStateRegistry.sol
gpersoon: that would be checking twice, so no need to add

PR #519, opened 2024-05-08
chore: add layerzero v2 

sujithsomraaj:


src/crosschain-data/adapters/layerzero/LayerzeroImplementation.sol
```diff
@@ -19,7 +19,7 @@ contract LayerzeroImplementation is IAmbImplementation, ILayerZeroUserApplicatio
     using DataLib for uint256;
 
     //////////////////////////////////////////////////////////////
```
0xTimepunk, line 21: This change would affect the bytecode

src/interfaces/IAmbImplementation.sol
```diff
@@ -5,6 +5,13 @@ pragma solidity ^0.8.23;
 /// @dev Interface for arbitrary message bridge (AMB) implementations
 /// @author ZeroPoint Labs
 interface IAmbImplementation {
+     //////////////////////////////////////////////////////////////
```
0xTimepunk, line 8: Shouldn't make this change as it will affect bytecode (however we're are redeploying all ambs I guess, so maybe not a problem)

src/crosschain-data/adapters/layerzero/LayerzeroImplementation.sol
```diff
@@ -19,7 +19,7 @@ contract LayerzeroImplementation is IAmbImplementation, ILayerZeroUserApplicatio
...
-    //                         CONSTANTS                         //
+    //                         CONSTANTS                        //
```
0xTimepunk, line 22: This change would affect the bytecode

test/unit/crosschain-data/adapters/LayerzeroV2Implementation.t.sol
```diff
@@ -0,0 +1,226 @@
+// SPDX-License-Identifier: MIT
```
0xTimepunk, line 1: Do we really need to preserve the tests for the old one?

test/unit/crosschain-liquidity/DstSwapper.t.sol
```diff
@@ -182,7 +182,7 @@ contract DstSwapperTest is ProtocolActions {
         );
 
         vm.startPrank(deployer);
-        deal(getContract(ETH, "WETH"), dstSwapper, 1e18);
+        deal(getContract(ETH, "WETH"), dstSwapper, 100e18);
```
0xTimepunk, line 185: what happened that we need to start doing this?

test/utils/BaseSetup.sol
```diff
@@ -185,9 +186,9 @@ abstract contract BaseSetup is DSTest, StdInvariant, Test {
     /// @notice id 2 is hyperlane
     /// @notice id 3 is wormhole (Automatic Relayer)
     /// @notice id 4 is wormhole (Specialized Relayer)
-
-    uint8[] public ambIds = [uint8(1), 2, 3, 4];
-    bool[] public isBroadcastAMB = [false, false, false, true];
+    /// @notice id 6 is layerzero-v2
```
0xTimepunk, line 189: could re-use 1 for the purpose of the test, but I guess you decided to keep LZ v1 in the tests

0xTimepunk (commented):
Some general comments but looks good. I don't like the idea of adding all those files to the vendor folder, because technically we modified OApp.sol so it would need to be audited?

src/crosschain-data/adapters/layerzero-v2/LayerzeroV2Implementation.sol
```diff
@@ -0,0 +1,255 @@
...
+    //////////////////////////////////////////////////////////////
+
+    event EndpointUpdated(address indexed oldEndpoint_, address indexed newEndpoint_);
+    event DelegateUpdated(address indexed newDelegate_);
+
+    //////////////////////////////////////////////////////////////
+    //                       MODIFIERS                          //
+    //////////////////////////////////////////////////////////////
+
+    modifier onlyProtocolAdmin() {
```
0xTimepunk, line 61: Try to use OApp from the lib folder (without remappings)

Override _checkOwner internal function to do PROTOCOL ADMIN CHECK

src/crosschain-data/adapters/layerzero-v2/LayerzeroV2Implementation.sol
```diff
@@ -0,0 +1,367 @@
...
+        if (peer == bytes32(0)) revert PEER_NOT_SET();
+        return peer;
+    }
+
+    /// @dev interacts with EndpointV2.quote() for fee calculation
+    function _quote(
+        uint32 _dstEid,
+        bytes memory _message,
+        bytes memory _options,
+        bool _payInLzToken
```
gpersoon, line 285: `payInLzToken` could also be hardcoded to be false, because `_lzSend()` also does that

src/crosschain-data/adapters/layerzero-v2/LayerzeroV2Implementation.sol
```diff
@@ -0,0 +1,367 @@
...
+        uint64 dstChainId_,
+        bytes memory message_,
+        bytes memory extraData_
+    )
+        external
+        payable
+        override
+        onlyValidStateRegistry
+    {
+        uint32 eid = ambChainId[dstChainId_];
```
gpersoon, line 171: maybe check `endpoint` is set

src/crosschain-data/adapters/layerzero-v2/LayerzeroV2Implementation.sol
```diff
@@ -0,0 +1,367 @@
...
+        if (address(endpoint) != msg.sender) revert Error.CALLER_NOT_ENDPOINT();
+
+        /// @dev validates is source sender is valid
+        if (_getPeerOrRevert(origin_.srcEid) != origin_.sender) revert Error.INVALID_SRC_SENDER();
+
+        _lzReceive(origin_, guid_, message_, executor_, extraData_);
+    }
+
+    /// @inheritdoc IAmbImplementation
+    function retryPayload(bytes memory data_) external payable override { }
```
gpersoon, line 203: Maybe this is useful here: https://docs.layerzero.network/v2/developers/evm/troubleshooting/debugging-messages#retry-message

src/crosschain-data/adapters/layerzero-v2/LayerzeroV2Implementation.sol
```diff
@@ -0,0 +1,367 @@
...
+        /// extraData_ here is the layerzero options
+        fees = _quote(tempAmbChainId, message_, extraData_, false).nativeFee;
+    }
+
+    /// @inheritdoc IAmbImplementation
+    function generateExtraData(uint256 gasLimit) external pure override returns (bytes memory extraData) {
+        /// generate the executor options here, since we don't use msg.value just returning encoded args
+        /// refer: https://docs.layerzero.network/v2/developers/evm/gas-settings/options#lzreceive-option
+        return abi.encodePacked(OPTIONS_TYPE, gasLimit);
+    }
```
gpersoon, line 236:
```solidity
uint16 private constant OPTIONS_TYPE = 1;
```
==> seems to be `OPTION_TYPE_LZRECEIVE`, however that is an `uint8`:
https://github.com/LayerZero-Labs/LayerZero-v2/blob/main/protocol/contracts/messagelib/libs/ExecutorOptions.sol#L12
```solidity
 uint8 internal constant OPTION_TYPE_LZRECEIVE = 1;
```

`encodeLzReceiveOption` uses uint128, while `generateExtraData` uses uint256
https://github.com/LayerZero-Labs/LayerZero-v2/blob/main/oapp/contracts/oapp/libs/OptionsBuilder.sol#L53-L60
```solidity
  function encodeLzReceiveOption(uint128 _gas, uint128 _value) internal pure returns (bytes memory) {
        return _value == 0 ? abi.encodePacked(_gas) : abi.encodePacked(_gas, _value);
    }
```

==> these differences result in differences in the generated `bytes`

src/crosschain-data/adapters/layerzero-v2/LayerzeroV2Implementation.sol
gpersoon: okido

src/crosschain-data/adapters/layerzero-v2/LayerzeroV2Implementation.sol
gpersoon: ok, looks good
You could maybe add a comment that the legecy option is used and point to `encodeLegacyOptionsType1()` or even use the library

PR #532, opened 2024-05-22
feat: ERC5115 PoC

0xTimepunk:


src/forms/ERC5115Form.sol
```diff
@@ -0,0 +1,648 @@
...
+    }
+
+    function getRewardIndexesStored() public view virtual returns (uint256[] memory indexes) {
+        indexes = IStandardizedYield(vault).rewardIndexesStored();
+    }
+
+    function getRewardTokens() public view virtual returns (address[] memory rewardTokens) {
+        rewardTokens = IStandardizedYield(vault).getRewardTokens();
+    }
+
```
vikramarun, line 171: is this the same as .asset()?

src/forms/ERC5115Form.sol
```diff
@@ -0,0 +1,648 @@
...
+    function superformYieldTokenSymbol() external view virtual override returns (string memory) {
+        return string(abi.encodePacked("sp-", IERC20Metadata(vault).symbol()));
+    }
+
+    /// @inheritdoc BaseForm
+    function getStateRegistryId() external view override returns (uint8) {
+        return STATE_REGISTRY_ID;
+    }
+
+    function getExchangeRate() public view virtual returns (uint256 exchangeRate) {
```
vikramarun, line 156: propose that we call this getPricePerVaultShare instead of renaming to getExchangeRate, this way we can reliably call a given function on any Superform to get the price

src/forms/ERC5115Form.sol
```diff
@@ -0,0 +1,648 @@
...
+    function getVaultSymbol() public view virtual override returns (string memory) {
+        return IStandardizedYield(vault).symbol();
+    }
+
+    /// @inheritdoc BaseForm
+    function getVaultDecimals() public view virtual override returns (uint256) {
+        return uint256(IStandardizedYield(vault).decimals());
+    }
+
+    /// @inheritdoc BaseForm
```
vikramarun, line 101: think we should use exchangeRate

src/forms/ERC5115Form.sol
```diff
@@ -0,0 +1,648 @@
...
+        return 0;
+    }
+
+    /// @inheritdoc BaseForm
+    function getVaultShareBalance() public view virtual override returns (uint256) {
+        return IStandardizedYield(vault).balanceOf(address(this));
+    }
+
+    /// @inheritdoc BaseForm
+    function getTotalAssets() public view virtual override returns (uint256) {
```
vikramarun, line 112: how do we get the TVL of a vault in 5115?

src/forms/ERC5115Form.sol
```diff
@@ -0,0 +1,648 @@
...
+        return IStandardizedYield(vault).symbol();
+    }
+
+    /// @inheritdoc BaseForm
+    function getVaultDecimals() public view virtual override returns (uint256) {
+        return uint256(IStandardizedYield(vault).decimals());
+    }
+
+    /// @inheritdoc BaseForm
+    function getPricePerVaultShare() public view virtual override returns (uint256) {
```
vikramarun, line 102: we could implement this as   `price = IStandardizedYield(vault).exchangeRate();`

src/forms/ERC5115Form.sol
vikramarun, line 112: how do we get total TVL of a vault in 5115?

src/forms/ERC5115Form.sol
```diff
@@ -0,0 +1,648 @@
...
+
+    function getRewardIndexesStored() public view virtual returns (uint256[] memory indexes) {
+        indexes = IStandardizedYield(vault).rewardIndexesStored();
+    }
+
+    function getRewardTokens() public view virtual returns (address[] memory rewardTokens) {
+        rewardTokens = IStandardizedYield(vault).getRewardTokens();
+    }
+
+    function getYieldToken() public view virtual returns (address yieldToken) {
```
vikramarun, line 172: is this .asset()?

src/forms/ERC5115Form.sol
0xTimepunk: Done. Removed getExchangeRate in favor of this

src/forms/ERC5115Form.sol
0xTimepunk: "This read-only method returns the underlying yield-bearing token (representing a GYGP) address.

MUST return a token address that conforms to the ERC-20 interface, or zero address
MUST NOT revert.
MUST reflect the exact underlying yield-bearing token address if the SY token is a wrapped token.
MAY return 0x or zero address if the SY token is natively implemented, and not from wrapping."

src/forms/ERC5115Form.sol
0xTimepunk: Since 5115 can be multiple assets there isn't a direct totalAssets() functionality. I think this could be approximated via grabbing the balance of each address of getTokensIn() ?

src/forms/ERC5115Form.sol
```diff
@@ -0,0 +1,714 @@
...
+    }
+
+    /// @inheritdoc BaseForm
+    function getVaultShareBalance() public view virtual override returns (uint256) {
+        return IStandardizedYield(vault).balanceOf(address(this));
+    }
+
+    /// @inheritdoc BaseForm
+    function getTotalAssets() public view virtual override returns (uint256) {
+        return 0;
```
gpersoon, line 118: maybe get an approximation with a calculation based on `exchangeRate()` and `totalSupply()` ?

src/forms/ERC5115Form.sol
```diff
@@ -0,0 +1,714 @@
...
+    }
+
+    /// @inheritdoc BaseForm
+    function getTotalSupply() public view virtual override returns (uint256) {
+        return IERC20Metadata(vault).totalSupply();
+    }
+
+    /// @inheritdoc BaseForm
+    function getPreviewPricePerVaultShare() public view virtual override returns (uint256) {
+        return 0;
```
gpersoon, line 128: could be implemented in the same way as `getPricePerVaultShare()`?

src/forms/ERC5115Form.sol
```diff
@@ -0,0 +1,714 @@
...
+    }
+
+    /// @inheritdoc BaseForm
+    function getStateRegistryId() external pure override returns (uint8) {
+        return stateRegistryId;
+    }
+
+    /// @inheritdoc IERC5115Form
+    function getAccruedRewards(address user) public view virtual override returns (uint256[] memory rewards) {
+        rewards = IStandardizedYield(vault).accruedRewards(user);
```
gpersoon, line 163: What if the ERC5115 token doesn't implement `accruedRewards()`?
Same for `rewardIndexesStored()`, `getRewardTokens()`

gpersoon, line 163: How will `claimRewards()` be called?

src/forms/ERC5115Form.sol
```diff
@@ -0,0 +1,714 @@
...
+        view
+        virtual
+        returns (address[] memory tokensOut, uint256[] memory balances)
+    {
+        tokensOut = IStandardizedYield(vault).getTokensOut();
+        uint256 len = tokensOut.length;
+
+        balances = new uint256[](len);
+        for (uint256 i = 0; i < len; i++) {
+            try IERC20(tokensOut[i]).balanceOf(vault) returns (uint256 balance) {
```
gpersoon, line 213: This can still revert at the decoding of the return value e.g. if the `balanceOf()` doesn't return a `uint256`  (for example when the contract doesn't exist)

src/forms/ERC5115Form.sol
```diff
@@ -0,0 +1,714 @@
...
+                balances[i] = balance;
+            } catch {
+                balances[i] = address(vault).balance;
```
gpersoon, line 216: This retrieves the ETH balance, every time there is an error with balanceOf, this doesn't seem right

src/forms/ERC5115Form.sol
```diff
@@ -0,0 +1,714 @@
...
+        /// @dev pulling from sender, to auto-send tokens back in case of failed deposits / reverts
+        IERC20(vaultTokenIn).safeTransferFrom(msg.sender, address(this), singleVaultData_.amount);
+
+        /// @dev allowance is modified inside of the IERC20.transferFrom() call
+        IERC20(vaultTokenIn).safeIncreaseAllowance(vaultLoc, singleVaultData_.amount);
+
+        /// @dev deposit vaultTokenIn for shares and add extra validation check to ensure intended ERC5115 behavior
+        shares = _depositAndValidate(singleVaultData_, singleVaultData_.amount, vaultTokenIn);
+
+        emit Processed(srcChainId_, dstChainId, singleVaultData_.payloadId, singleVaultData_.amount, vaultLoc);
```
gpersoon, line 462: maybe include `vaultTokenIn` in the `emit()` (otherwise difficult to understand `amount`)

src/forms/wrappers/ERC5115To4626Wrapper.sol
```diff
@@ -0,0 +1,225 @@
...
+        return amountSharesOut;
+    }
+
+    /// @inheritdoc IStandardizedYield
+    function redeem(
+        address receiver,
+        uint256 amountSharesToRedeem,
+        address tokenOut,
+        uint256 minTokenOut,
+        bool /*burnFromInternalBalance*/
```
gpersoon, line 82: I think `burnFromInternalBalance` requires some additional code to implement (can't just ignore it)

src/forms/ERC5115Form.sol
```diff
@@ -0,0 +1,714 @@
...
+import { IERC20Metadata } from "openzeppelin-contracts/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+import { SafeERC20 } from "openzeppelin-contracts/contracts/token/ERC20/utils/SafeERC20.sol";
+import { IStandardizedYield } from "src/vendor/pendle/IStandardizedYield.sol";
+
+/// @title ERC5115Form
+/// @dev The Form implementation for ERC5115 vaults
+/// @notice Reference implementation of a vault:
+/// https://github.com/pendle-finance/pendle-core-v2-public/blob/main/contracts/core/StandardizedYield/SYBase.sol
+/// @author Zeropoint Labs
+contract ERC5115Form is IERC5115Form, BaseForm, LiquidityHandler {
```
gpersoon, line 21: There is a lot of overlap with ERC4626FormImplementation, maybe useful to refactor both into one code base for easier maintenance

src/forms/ERC5115Form.sol
0xTimepunk: To give more context to this:
Due to the fact that our factory reads asset() when deploying superform and this function is not available in ERC5115  https://github.com/superform-xyz/superform-core/blob/b95ba8568854ce05271c6b7502346f008fa7f922/src/SuperformFactory.sol#L238 (our architecture wasn't built for allowing multiple asset in in a straightforward way), we had to build a wrapper contract to 5115 adding the 4626 asset() logic. This requires setting a tokenIn in the constructor (so a give 5115 will have multiple wrappers for the different asset in) https://github.com/superform-xyz/superform-core/blob/draftERC5115/src/forms/wrappers/ERC5115To4626Wrapper.sol

With this change we can now calculate total assets from the perspective of the wrapper 

https://github.com/superform-xyz/superform-core/blob/b95ba8568854ce05271c6b7502346f008fa7f922/src/forms/ERC5115Form.sol#L119-L121

src/forms/ERC5115Form.sol
0xTimepunk: I would say for previewWithdrawFrom wouldn't make sense to implement since withdraw function doesn't exist in 5115.

PreviewRedeemFrom is now implemented because the form is now aware of asset (one of the tokenIn),.

src/forms/ERC5115Form.sol
0xTimepunk: Still to be implemented! Will add an update on this

0xTimepunk: I think we could try catch this to provide better error handling since those functions are not part of the standard. Provided a fix

src/forms/ERC5115Form.sol
0xTimepunk: Unfortunately due to CREATE2 assumptions of preserving bytecode when we want to release the codebase to new chains we have to keep existing deployed code as it is to keep the same contract addresses (albeit not too critical for the implementation contract, cc @vikramarun )

src/forms/ERC5115Form.sol
0xTimepunk: Since asset (token In) is now tied to the form / vault wrapper, I believe this may not be necessary (see updated code)

src/forms/wrappers/ERC5115To4626Wrapper.sol
0xTimepunk: Makes sense, amended

src/forms/ERC5115Form.sol
0xTimepunk: I'll think a bit more about this as we're still wrapping up the redeem side of the form

src/forms/ERC5115Form.sol
0xTimepunk: Similar comment to above, will likely be refactored/removed as this idea may not make sense anymore

src/forms/ERC5115Form.sol
0xTimepunk: This function was deleted given setting main token out and avoiding all these problems

src/vendor/pendle/IStandardizedYield.sol
```diff
@@ -0,0 +1,180 @@
...
+     *
+     * Requirements:
+     * - (`tokenIn`) must be a valid base token.
+     */
+    function deposit(
+        address receiver,
+        address tokenIn,
+        uint256 amountTokenToDeposit,
+        uint256 minSharesOut
+    )
```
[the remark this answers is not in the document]
0xTimepunk: Checking with their team

src/forms/ERC5115Form.sol
```diff
@@ -0,0 +1,758 @@
...
+    function getStateRegistryId() external pure override returns (uint8) {
+        return stateRegistryId;
+    }
+
+    /// @inheritdoc IERC5115Form
+    function getAccruedRewards(address user) public view virtual override returns (uint256[] memory) {
+        try IStandardizedYield(vault).accruedRewards(user) returns (uint256[] memory rewards) {
+            return rewards;
+        } catch {
+            revert FUNCTION_NOT_IMPLEMENTED();
```
gpersoon, line 171: could also return an array of 0's, not sure if this would be better
(similar for `getRewardIndexesStored` and `getRewardTokens`)

src/forms/ERC5115Form.sol
```diff
@@ -0,0 +1,758 @@
...
+
+    function _processEmergencyWithdraw(address receiverAddress_, uint256 amount_) internal {
+        IStandardizedYield v = IStandardizedYield(vault);
+        if (receiverAddress_ == address(0)) revert Error.ZERO_ADDRESS();
+
+        if (v.balanceOf(address(this)) < amount_) {
+            revert Error.INSUFFICIENT_BALANCE();
+        }
+
+        v.safeTransfer(receiverAddress_, amount_);
```
gpersoon, line 741: maybe transfer the assets from the underlying vault?
(requires additional code in ERC5115To4626Wrapper i guess)

src/forms/wrappers/ERC5115To4626Wrapper.sol
```diff
@@ -0,0 +1,320 @@
...
+        /// @dev tokenIn is forwarded to this wrapper first
+        _transferIn(tokenIn, msg.sender, amountTokenToDeposit);
+
+        /// @dev allowance is increased for non native tokens
+        if (tokenIn != NATIVE) {
+            IERC20(tokenIn).safeIncreaseAllowance(vault, amountTokenToDeposit);
+        }
+
+        /// @dev deposit is made and any allowance is reset
+        amountSharesOut = IStandardizedYield(vault).deposit(receiver, tokenIn, amountTokenToDeposit, minSharesOut);
```
gpersoon, line 125: if `tokenIn == NATIVE`, shouldn't it be send to deposit?  ({value: ...} )

src/forms/ERC5115Form.sol
```diff
@@ -0,0 +1,758 @@
...
+        (uint256 nVaults, bytes memory extra5115Data) = abi.decode(singleVaultData_.extraFormData, (uint256, bytes));
+
+        bytes memory recursion5115Data;
+        address vaultTokenIn;
+        uint256 superformId;
+        bool found5115;
+
+        for (uint256 i = 0; i < nVaults; ++i) {
+            (recursion5115Data, superformId, vaultTokenIn) =
+                abi.decode(i == 0 ? extra5115Data : recursion5115Data, (bytes, uint256, address));
```
gpersoon, line 423: `recursion5115Data` isn't really necessary I think, could also be:
```solidity
 (extra5115Data, superformId, vaultTokenIn) = abi.decode(extra5115Data, (bytes, uint256, address));
```

src/forms/wrappers/ERC5115To4626Wrapper.sol
```diff
@@ -0,0 +1,320 @@
...
+        if (receiver == address(this)) revert INVALID_RECEIVER();
+
+        /// @dev the share is transfered to this contract first
+        IERC20(vault).safeTransferFrom(msg.sender, address(this), amountSharesToRedeem);
+
+        /// @dev redeem will burn the share from this contract
+        return IStandardizedYield(vault).redeem(
+            receiver, amountSharesToRedeem, tokenOut, minTokenOut, burnFromInternalBalance
+        );
+    }
```
gpersoon, line 156: After checking [redeem from SYBase](https://github.com/pendle-finance/pendle-core-v2-public/blob/main/contracts/core/StandardizedYield/SYBase.sol#L59-L78) and thinking about this a bit more, I think the code should be:
```solidity
if (burnFromInternalBalance == false) {
    IERC20(vault).safeTransferFrom(msg.sender, address(this), amountSharesToRedeem);
}
return IStandardizedYield(vault).redeem(..., false); // burns from this
```

src/forms/ERC5115Form.sol
0xTimepunk: Unfortunately it is required because ERC5115Forms (and forms in general) are not aware of multi vault cases. And in this situation, due to a shortsight in how extraFormData is passed in https://github.com/superform-xyz/superform-core/blob/0bf540f8c9b46f2da487ccaf6f9d5eeb5ac06bd7/src/crosschain-data/extensions/CoreStateRegistry.sol#L815 , this is the only way to pass in data for such case

src/forms/wrappers/ERC5115To4626Wrapper.sol
0xTimepunk: Native handling is disallowed in form and CoreStateRegistry, so did not include this case

src/vendor/pendle/IStandardizedYield.sol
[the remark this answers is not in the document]
0xTimepunk: @gpersoon , reply from the team

"You can rely on the IStandardYield interface instead. We will be updating the EIP to reflect this as we're also still collecting feedback around. But you can consider the interface in our repo which is being actively used as the canonical one"

src/forms/ERC5115Form.sol
0xTimepunk: The goal of this function is to transfer the shares that the ERC5115Form is holding. These shares belong to the underlying vault, obtained with `IERC5115To4626Wrapper(vault).getUnderlying5115Vault()`.

Substituted v here to account for that

src/forms/wrappers/ERC5115To4626Wrapper.sol
```diff
@@ -0,0 +1,320 @@
...
+    address public immutable mainTokenOut;
+
+    address internal constant NATIVE = address(0);
+
+    //////////////////////////////////////////////////////////////
+    //                      MODIFIER                         //
+    //////////////////////////////////////////////////////////////
+
+    /// @dev these modifiers help protect against undesired changes of tokenIn or out
+    modifier validateTokenIn() {
```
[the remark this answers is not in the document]
0xTimepunk: Fixed

src/forms/wrappers/ERC5115To4626Wrapper.sol
0xTimepunk: Assuming you meant

```
if (burnFromInternalBalance == false) {
    IERC20(vault).safeTransferFrom(msg.sender, address(this), amountSharesToRedeem);
}
 /// @dev redeem will burn the share from this contract
return IStandardizedYield(vault).redeem(..., burnFromInternalBalance); // burns from this
```

src/forms/ERC5115Form.sol
0xTimepunk: I think an explicit error in case the function is not implemented could be easier to interpret for off chain actors

PR #561, opened 2024-06-11
feat: ERC5115Form

0xTimepunk:


src/forms/wrappers/ERC5115To4626Wrapper.sol
```diff
@@ -0,0 +1,309 @@
...
+
+        /// @dev receiver cannot be this wrapper contract
+        if (receiver == address(this)) revert INVALID_RECEIVER();
+
+        if (burnFromInternalBalance == false) {
+            IERC20(vault).safeTransferFrom(msg.sender, address(this), amountSharesToRedeem);
+        }
+
+        return IStandardizedYield(vault).redeem(
+            receiver, amountSharesToRedeem, tokenOut, minTokenOut, burnFromInternalBalance
```
gpersoon, line 131: Due to the double wrapping I think burnFromInternalBalance should not be used but false. 
Please double check, perhaps with a test

src/forms/ERC5115Form.sol
```diff
@@ -0,0 +1,703 @@
...
+        (,, uint64 dstChainId) = singleVaultData_.superformId.getSuperform();
+        address vaultLoc = vault;
+
+        /// @dev for deposits tokenIn must be decoded from extraFormData as interimToken may be in use
+        /// @dev Warning: This must be validated by a keeper to be the token received in CSR for the given payload, as
+        /// this can be forged by the user
+        /// @dev and it's not possible to validate on chain the final token post bridging/swapping
+        (uint256 nVaults, bytes memory extra5115Data) = abi.decode(singleVaultData_.extraFormData, (uint256, bytes));
+
+        bytes memory recursion5115Data;
```
gpersoon, line 410: I meant the extra variabele recursion5115Data isn't necessary, but could be integrated with extra5115Data
(Small suggestion)

0xTimepunk: You are right

PR #562, opened 2024-06-11
feat: introduce router plus [SUP-7853, SUP-7876, SUP-7873, SUP-7878, SUP-7815, SUP-8098, SUP-8102, SUP-8124]

sujithsomraaj:


src/router-plus/BaseSuperformRouterPlus.sol
```diff
@@ -0,0 +1,136 @@
...
+    function _callSuperformRouter(bytes memory callData_, uint256 msgValue_) internal {
+        (bool success, bytes memory returndata) =
+            _getAddress(keccak256("SUPERFORM_ROUTER")).call{ value: msgValue_ }(callData_);
+
+        Address.verifyCallResult(success, returndata);
+    }
+
+    function _deposit(IERC20 asset_, uint256 amountToDeposit_, uint256 msgValue_, bytes memory callData_) internal {
+        /// @dev approves superform router on demand
+        asset_.approve(_getAddress(keccak256("SUPERFORM_ROUTER")), amountToDeposit_);
```
gpersoon, line 123: for certain tokens its important to first approve with 0

src/router-plus/BaseSuperformRouterPlus.sol
```diff
@@ -0,0 +1,136 @@
...
+
+        /// TODO confirm if we should validate depositCallData with receiverAddressSP and other parameters?
+        /// @notice this is used in all actions. In cross chain rebalances, most key data is validated, but not on
+        /// @notice same chain rebalances or deposits
+
+        _callSuperformRouter(callData_, msgValue_);
+    }
```
gpersoon, line 130: could potentially reset any remaining approval

src/router-plus/SuperformRouterPlus.sol
```diff
@@ -0,0 +1,503 @@
...
+
+    function _transferBatchSuperPositions(address user_, uint256[] memory ids_, uint256[] memory amounts_) internal {
+        address superPositions = _getAddress(keccak256("SUPER_POSITIONS"));
+
+        SuperPositions(superPositions).safeBatchTransferFrom(user_, address(this), ids_, amounts_, "");
+        SuperPositions(superPositions).setApprovalForAll(_getAddress(keccak256("SUPERFORM_ROUTER")), true);
+    }
+
+    function _transferERC20In(IERC20 erc20_, address user_, uint256 amount_) internal {
+        erc20_.transferFrom(user_, address(this), amount_);
```
gpersoon, line 428: check results or use safeTransferFrom

src/router-plus/SuperformRouterPlus.sol
```diff
@@ -0,0 +1,503 @@
...
+        if (msg.value < totalFee) {
+            revert INVALID_FEE();
+        }
+    }
+
+    function _tokenRefunds(address asset_, address user_, uint256 balanceBefore) internal {
+        uint256 balanceDiff = IERC20(asset_).balanceOf(address(this)) - balanceBefore;
+
+        if (balanceDiff > 0) {
+            IERC20(asset_).transfer(user_, balanceDiff);
```
gpersoon, line 482: check results or use safeTransfer

src/router-plus/SuperformRouterPlus.sol
```diff
@@ -0,0 +1,503 @@
...
+        }
+
+        if (IERC20(asset_).allowance(address(this), _getAddress(keccak256("SUPERFORM_ROUTER"))) > 0) {
+            IERC20(asset_).forceApprove(_getAddress(keccak256("SUPERFORM_ROUTER")), 0);
+        }
+    }
+
+    /// @dev refunds any unused refunds
```
gpersoon, line 490: comment is a bit weird (refunds mentioned twice)

src/router-plus/SuperformRouterPlus.sol
```diff
@@ -0,0 +1,503 @@
...
+        /// @notice msg.value here is the sum of rebalanceFromMsgValue and rebalanceToMsgValue (to be executed later by
+        /// the keeper)
+        _callSuperformRouter(args.callData, msg.value);
+
+        if (!whitelistedSelectors[Actions.DEPOSIT][args.rebalanceToSelector]) {
+            revert INVALID_DEPOSIT_SELECTOR();
+        }
+        ++ROUTER_PLUS_PAYLOAD_ID;
+
+        uint256 routerPlusPayloadId = ROUTER_PLUS_PAYLOAD_ID;
```
gpersoon, line 158: can be done in one operation so safe some gas:
```solidity
uint256 routerPlusPayloadId = ++ROUTER_PLUS_PAYLOAD_ID;
```

src/router-plus/SuperformRouterPlusAsync.sol
```diff
@@ -0,0 +1,406 @@
...
+    function setXChainRebalanceCallData(
+        address receiverAddressSP_,
+        uint256 routerPlusPayloadId_,
+        XChainRebalanceData memory data_
+    )
+        external
+        override
+        onlyRouterPlus
+    {
+        xChainRebalanceCallData[receiverAddressSP_][routerPlusPayloadId_] = data_;
```
gpersoon, line 92: could check the value of `xChainRebalanceCallData` for the `routerPlusPayloadId_` hasn't been set yet.
this might be relevant with a new version of `SuperformRouterPlus` that restarts the `ROUTER_PLUS_PAYLOAD_ID`

src/router-plus/SuperformRouterPlusAsync.sol
```diff
@@ -0,0 +1,406 @@
...
+    function getXChainRebalanceCallData(
+        address receiverAddressSP_,
+        uint256 routerPlusPayloadId_
+    )
+        external
+        view
+        override
+        returns (XChainRebalanceData memory)
+    {
+        return xChainRebalanceCallData[receiverAddressSP_][routerPlusPayloadId_];
```
gpersoon, line 75: `xChainRebalanceCallData[][]` is already public so getter function not really necessary.
Alternatively `xChainRebalanceCallData[][]` could be made internal

src/router-plus/SuperformRouterPlus.sol
```diff
@@ -0,0 +1,503 @@
...
+        address superPositions = _getAddress(keccak256("SUPER_POSITIONS"));
+        SuperPositions(superPositions).safeTransferFrom(user_, address(this), id_, amount_, "");
+        SuperPositions(superPositions).setApprovalForOne(_getAddress(keccak256("SUPERFORM_ROUTER")), id_, amount_);
+    }
+
+    function _transferBatchSuperPositions(address user_, uint256[] memory ids_, uint256[] memory amounts_) internal {
+        address superPositions = _getAddress(keccak256("SUPER_POSITIONS"));
+
+        SuperPositions(superPositions).safeBatchTransferFrom(user_, address(this), ids_, amounts_, "");
+        SuperPositions(superPositions).setApprovalForAll(_getAddress(keccak256("SUPERFORM_ROUTER")), true);
```
gpersoon, line 424: This sets the approval for all id's which quite is a lot
Does only have to be done once (however only as long as the SUPERFORM_ROUTER stays the same)
Perhaps reset it when you are done to limit any potential risk

src/router-plus/BaseSuperformRouterPlus.sol
```diff
@@ -0,0 +1,136 @@
+// SPDX-License-Identifier: BUSL-1.1
```
gpersoon, line 1: maybe add `forwardDustToPaymaster()` ?

src/router-plus/SuperformRouterPlusAsync.sol
```diff
@@ -0,0 +1,406 @@
...
+        return delay;
+    }
+
+    function _castToMultiVaultData(SingleVaultSFData memory data_)
+        internal
+        pure
+        returns (MultiVaultSFData memory castedData_)
+    {
+        uint256[] memory superformIds = new uint256[](1);
+        superformIds[0] = data_.superformId;
```
gpersoon, line 326: Could have a function `castUint256ToArray()`
(could also be added to & used in ArrayCastLib)
(could also be used in rebalanceSinglePosition for sharesToRedeem)

src/router-plus/SuperformRouterPlus.sol
```diff
@@ -0,0 +1,503 @@
...
+            if (req.superformData.liqRequest.token != args.asset) {
+                revert REBALANCE_SINGLE_POSITIONS_DIFFERENT_TOKEN();
+            }
+            if (req.superformData.liqRequest.liqDstChainId != CHAIN_ID) {
+                revert REBALANCE_SINGLE_POSITIONS_DIFFERENT_CHAIN();
+            }
+            if (req.superformData.amount != args.sharesToRedeem[0]) {
+                revert REBALANCE_SINGLE_POSITIONS_DIFFERENT_AMOUNT();
+            }
+        } else {
```
gpersoon, line 376: could add comment: then must be Actions.REBALANCE_FROM_MULTI

src/router-plus/SuperformRouterPlus.sol
```diff
@@ -0,0 +1,503 @@
...
+
+            if (req.superformData.liqRequest.token != args.asset) {
+                revert REBALANCE_SINGLE_POSITIONS_DIFFERENT_TOKEN();
+            }
+            if (req.superformData.liqRequest.liqDstChainId != CHAIN_ID) {
+                revert REBALANCE_SINGLE_POSITIONS_DIFFERENT_CHAIN();
+            }
+            if (req.superformData.amount != args.sharesToRedeem[0]) {
+                revert REBALANCE_SINGLE_POSITIONS_DIFFERENT_AMOUNT();
+            }
```
gpersoon, line 375: note: receiverAddress and retain4626 are not checked, probably no problem because the
checks amountToDeposit / slippage will detect it

src/router-plus/SuperformRouterPlus.sol
```diff
@@ -0,0 +1,503 @@
...
+        }
+    }
+
+    /// @dev refunds any unused refunds
+    function _refundUnused(address asset_, address user_, uint256 balanceBefore, uint256 totalFee) internal {
+        _tokenRefunds(asset_, user_, balanceBefore);
+
+        if (msg.value > totalFee) {
+            /// @dev refunds msg.sender if msg.value was more than needed
+            (bool success,) = payable(msg.sender).call{ value: msg.value - totalFee }("");
```
gpersoon, line 496: note: via the `call()` a potential reentrancy can be done, but no problem because `_refundUnused()` is called at the end

src/router-plus/SuperformRouterPlus.sol
```diff
@@ -0,0 +1,503 @@
...
+        if (req.superformData.amount != args.sharesToRedeem) {
+            revert REBALANCE_SINGLE_POSITIONS_DIFFERENT_AMOUNT();
+        }
+
+        address ROUTER_PLUS_ASYNC = _getAddress(keccak256("SUPERFORM_ROUTER_PLUS_ASYNC"));
+
+        if (req.superformData.receiverAddress != ROUTER_PLUS_ASYNC) {
+            revert REBALANCE_XCHAIN_INVALID_RECEIVER_ADDRESS();
+        }
+
```
gpersoon, line 147: note:  _parseSelectorMem(args.callData) (selector) now can only be IBaseRouter.singleXChainSingleVaultWithdraw.selector, which is due to the whitelistedSelectors check.
perhaps add this as a comment for the reader

src/router-plus/SuperformRouterPlus.sol
```diff
@@ -0,0 +1,503 @@
...
+            for (uint256 i; i < len; ++i) {
+                uint256 len2 = req.superformsData[i].liqRequests.length;
+
+                for (uint256 j; j < len2; ++j) {
+                    // Validate that the token and chainId is equal in all indexes
+                    if (req.superformsData[i].liqRequests[j].token != args.interimAsset) {
+                        revert REBALANCE_MULTI_POSITIONS_DIFFERENT_TOKEN();
+                    }
+                    if (req.superformsData[i].liqRequests[j].liqDstChainId != CHAIN_ID) {
```
gpersoon, line 248: If it is always the same chain it isn't really Multi Destionation I think
So maybe not worth the trouble to implement code for multiDstMultiVaultWithdraw

src/router-plus/SuperformRouterPlus.sol
```diff
@@ -0,0 +1,503 @@
...
+                        revert REBALANCE_MULTI_POSITIONS_DIFFERENT_CHAIN();
+                    }
+                    if (req.superformsData[i].amounts[j] != args.sharesToRedeem[j]) {
```
gpersoon, line 251: This doesn't seem right sharesToRedeem[j] is used multiple times: for each i,
I would expect is also to be a 2 dimentional array

src/router-plus/SuperformRouterPlusAsync.sol
```diff
@@ -0,0 +1,406 @@
...
+
+            MultiVaultSFData memory multiSuperformData = _updateSuperformData(
+                _castToMultiVaultData(superformData), args_.liqRequests[0], args_.receiverAddressSP
+            );
+
+            superformData.liqRequest.txData = multiSuperformData.liqRequests[0].txData;
+            superformData.liqRequest.nativeAmount = multiSuperformData.liqRequests[0].nativeAmount;
+
+            rebalanceToCallData =
+                abi.encodeWithSelector(data.rebalanceSelector, SingleDirectSingleVaultStateReq(superformData));
```
gpersoon, line 165: Could also use `abi.encodeCall()` with the `real` function identifier and check that the selector of that function is indeed equal to `data.rebalanceSelector`
The advantage would be that `abi.encodeCall()` checks the parameters

src/router-plus/SuperformRouterPlus.sol
```diff
@@ -0,0 +1,503 @@
...
+        );
+    }
+
+    /// @inheritdoc ISuperformRouterPlus
+    function deposit4626(address vault_, Deposit4626Args calldata args) external payable override {
+        _transferERC20In(IERC20(vault_), args.receiverAddressSP, args.amount);
+        IERC4626 vault = IERC4626(vault_);
+        uint256 amountRedeemed = _redeemShare(vault, args.amount);
+
+        address assetAdr = vault.asset();
```
gpersoon, line 329: if vault is malicious, it can return a different value for `asset()` then in `_redeemShare()`
(for example in `_redeemShare()` is redeems a worthless asset and lateron  `asset()` returns the address of for example `usdc`)
That way could any tokens present in SuperformRouterPlus can be stolen

src/router-plus/SuperformRouterPlus.sol
```diff
@@ -0,0 +1,503 @@
...
+        IERC20 asset = IERC20(vault_.asset());
+        uint256 collateralBalanceBefore = asset.balanceOf(address(this));
+
+        /// @dev redeem the vault shares and receive collateral
+        vault_.redeem(amountToRedeem_, address(this), address(this));
+
+        /// @dev collateral balance after
+        uint256 collateralBalanceAfter = asset.balanceOf(address(this));
+
+        balanceDifference = collateralBalanceAfter - collateralBalanceBefore;
```
gpersoon, line 441: shouldn't there be a slippage check (e.g. make sure you have redeemed at least a certain amount)?

src/router-plus/SuperformRouterPlus.sol
```diff
@@ -0,0 +1,503 @@
...
+        uint256 amountRedeemed = _redeemShare(vault, args.amount);
+
+        address assetAdr = vault.asset();
+        IERC20 asset = IERC20(assetAdr);
+
+        if (!whitelistedSelectors[Actions.DEPOSIT][_parseSelectorMem(args.depositCallData)]) {
+            revert INVALID_DEPOSIT_SELECTOR();
+        }
+
+        uint256 balanceBefore = asset.balanceOf(address(this));
```
gpersoon, line 336: should't this balanceBefore be done before the redeem?
Now certainly after the deposit, the balance after is lower than the balance before and _tokenRefunds will revert
```solidity
function _tokenRefunds(address asset_, address user_, uint256 balanceBefore) internal {
    uint256 balanceDiff = IERC20(asset_).balanceOf(address(this)) - balanceBefore;
   ...
}     
```

src/router-plus/BaseSuperformRouterPlus.sol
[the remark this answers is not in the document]
0xTimepunk: fixed

0xTimepunk: Fixed

src/router-plus/SuperformRouterPlus.sol
0xTimepunk: fixed

src/router-plus/SuperformRouterPlus.sol
0xTimepunk: fixed

src/router-plus/SuperformRouterPlusAsync.sol
0xTimepunk: added

src/router-plus/SuperformRouterPlusAsync.sol
0xTimepunk: fixed

src/router-plus/SuperformRouterPlus.sol
0xTimepunk: it's now reset in _refundUnusedAndResetApprovals (previously _refundUnused)

src/router-plus/BaseSuperformRouterPlus.sol
0xTimepunk: I feel this is a risky thing to add to RouterPlusAsync (which would inherit from this), which receives funds in transit from xChain Rebalancing actions (very much like CoreStateRegistry and the reason we didn't add it there)

Added it for now just to RouterPlus

src/router-plus/SuperformRouterPlusAsync.sol
0xTimepunk: Removed the getter

src/router-plus/SuperformRouterPlus.sol
0xTimepunk: added

src/router-plus/SuperformRouterPlus.sol
0xTimepunk: Added a comment about this.

Also decided to a sanity check to make sure in the rebalance from step that receiverAddress is RouterPlus

src/router-plus/SuperformRouterPlus.sol
0xTimepunk: added

src/router-plus/SuperformRouterPlus.sol
0xTimepunk: I think I didn't completely understand this comment, but in a multicast call, some of the destinations can be same chain (which enters in the same chain flow). There's no prevention of this

src/router-plus/SuperformRouterPlus.sol
0xTimepunk: The way this array is organised has to do with the type of selector. I should have added a comment, but for multiDst all shares are organized in a single array. The array starts in the first destination with all the shares then it continues through all destinations with the same process

I substituted the j index through a counter that is incremented in each loop as that was indeed wrong

src/router-plus/SuperformRouterPlus.sol
0xTimepunk: Fixed

src/router-plus/SuperformRouterPlus.sol
0xTimepunk: added

src/router-plus/SuperformRouterPlus.sol
0xTimepunk: Fixed!

src/router-plus/SuperformRouterPlusAsync.sol
0xTimepunk: Added

src/router-plus/SuperformRouterPlusAsync.sol
0xTimepunk: Done!

src/router-plus/SuperformRouterPlus.sol
gpersoon: If I understand it correctly you can only do multiDstMultiVaultWithdraw to one chain?
(or maybe I misunderstood the check `liqDstChainId != CHAIN_ID` )

0xTimepunk:
```
    function multiDstMultiVaultWithdraw(MultiDstMultiVaultStateReq calldata req_)
        external
        payable
        override(BaseRouter, IBaseRouter)
    {
        uint256 balanceBefore = address(this).balance - msg.value;
        uint256 len = req_.dstChainIds.length;

        for (uint256 i; i < len; ++i) {
            if (CHAIN_ID == req_.dstChainIds[i]) {
                _singleDirectMultiVaultWithdraw(SingleDirectMultiVaultStateReq(req_.superformsData[i]));
            } else {
                _singleXChainMultiVaultWithdraw(
                    SingleXChainMultiVaultStateReq(req_.ambIds[i], req_.dstChainIds[i], req_.superformsData[i])
                );
            }
        }

        _forwardPayment(balanceBefore);
    }


```

So both same chain and dstChain are possible.

The check for liqDstChainId is just a sanity check to ensure when doing txData validation within the withdraw that the funds move back towards RouterPlus on CHAIN_ID

0xTimepunk (approved):

PR #563, opened 2024-06-13
chore: add claim rewards for 5115 form and increase coverage [SUP-7040, SUP-7255, SUP-7256]

sujithsomraaj:
**Notes on library update:** 
- `forge-std` added blast to default chain list


lib/forge-std
0xTimepunk, line 1: Can we revert these changes and use 1.8.2?

src/forms/ERC5115Form.sol
```diff
@@ -298,7 +322,24 @@ contract ERC5115Form is IERC5115Form, BaseForm, LiquidityHandler {
...
         /// @dev and it's not possible to validate on chain the final token post bridging/swapping
-        vars.vaultTokenIn = abi.decode(singleVaultData_.extraFormData, (address));
+        (uint256 nVaults, bytes memory extra5115Data) = abi.decode(singleVaultData_.extraFormData, (uint256, bytes));
+
+        uint256 superformId;
+        bool found5115;
+
+        for (uint256 i = 0; i < nVaults; ++i) {
+            (extra5115Data, superformId, vars.vaultTokenIn) =
+                abi.decode(i == 0 ? extra5115Data : extra5115Data, (bytes, uint256, address));
```
0xTimepunk, line 332: By removing the memory variable, this can now be `abi.decode( extra5115Data, (bytes, uint256, address));`

src/forms/ERC5115Form.sol
```diff
@@ -407,14 +448,13 @@ contract ERC5115Form is IERC5115Form, BaseForm, LiquidityHandler {
...
 
-        bytes memory recursion5115Data;
         address vaultTokenIn;
         uint256 superformId;
         bool found5115;
 
         for (uint256 i = 0; i < nVaults; ++i) {
-            (recursion5115Data, superformId, vaultTokenIn) =
-                abi.decode(i == 0 ? extra5115Data : recursion5115Data, (bytes, uint256, address));
+            (extra5115Data, superformId, vaultTokenIn) =
```
0xTimepunk, line 456: By removing the memory variable, this can now be abi.decode( extra5115Data, (bytes, uint256, address));

0xTimepunk (commented):
looks good to me, some final changes needed

src/forms/ERC5115Form.sol
```diff
@@ -184,6 +184,30 @@ contract ERC5115Form is IERC5115Form, BaseForm, LiquidityHandler {
         }
     }
 
+    /// @inheritdoc IERC5115Form
+    function claimRewardTokens() external virtual override {
+        address[] memory rewardTokens = getRewardTokens();
```
gpersoon, line 189: I was thinking: what if one of the reward tokens is a tokens that shouldn't be transferred out of the form contract?

src/forms/ERC5115Form.sol
```diff
@@ -184,6 +184,30 @@ contract ERC5115Form is IERC5115Form, BaseForm, LiquidityHandler {
         }
     }
 
+    /// @inheritdoc IERC5115Form
+    function claimRewardTokens() external virtual override {
```
gpersoon, line 188: Is it intentional that claimRewardTokens() isn't authorized?

src/forms/wrappers/ERC5115To4626Wrapper.sol
```diff
@@ -127,6 +127,7 @@ contract ERC5115To4626Wrapper is IERC5115To4626Wrapper {
             IERC20(vault).safeTransferFrom(msg.sender, address(this), amountSharesToRedeem);
         }
 
+        /// @dev ERC5115 form always uses `false` for burnFromInternalBalance
```
gpersoon, line 130: I still think the parameter to the underlying redeem should now be false  
(because it get the tokens from this contract and they are not transferred to the vault contract)

This is not an issue in the current code if burnFromInternalBalance is always false. But it could be an issue with future changes.

0xTimepunk (approved):

PR #614, opened 2024-09-04
chore: V1.5 featuring SuperformRouter Plus

0xTimepunk:


src/router-plus/SuperformRouterPlus.sol
```diff
@@ -0,0 +1,724 @@
...
+        address assetAdr = vault.asset();
+        IERC20 asset = IERC20(assetAdr);
+
+        uint256 balanceBefore = asset.balanceOf(address(this));
+
+        uint256 amountRedeemed = _redeemShare(vault, assetAdr, args.amount, args.expectedOutputAmount, args.maxSlippage);
+
+        uint256 amountIn = _validateAndGetAmountIn(args.depositCallData, amountRedeemed);
+
+        uint256 msgValue = msg.value / arrayLength;
```
gpersoon, line 642: this rounds down so the last bit of the msg.value isn't used

src/router-plus/SuperformRouterPlusAsync.sol
```diff
@@ -0,0 +1,558 @@
...
+    //                     STATE VARIABLES                      //
+    //////////////////////////////////////////////////////////////
+
+    mapping(address receiverAddressSP => mapping(uint256 routerPlusPayloadId => XChainRebalanceData data)) public
+        xChainRebalanceCallData;
+
+    mapping(uint256 routerPlusPayloadId => Refund) public refunds;
+    mapping(uint256 routerPlusPayloadId => bool processed) public processedRebalancePayload;
+
+    mapping(uint256 routerPlusPayloadId => bool approvedRefund) public approvedRefund;
```
gpersoon, line 39: gas optimization:
the mappings line 33-39 could potentially be (partly) combined because the first index parameter is the same.

especially `approvedRefund[]` could also be implemented by deleting `refunds[]`

src/router-plus/SuperformRouterPlusAsync.sol
```diff
@@ -0,0 +1,558 @@
...
+
+        XChainRebalanceData memory data = xChainRebalanceCallData[r.receiver][routerPlusPayloadId_];
+
+        if (requestedAmount > data.expectedAmountInterimAsset) {
+            revert REQUESTED_AMOUNT_TOO_HIGH();
+        }
+
+        refunds[routerPlusPayloadId_] = Refund(msg.sender, data.interimAsset, requestedAmount);
+
+        emit refundRequested(routerPlusPayloadId_, msg.sender, r.interimToken, requestedAmount);
```
gpersoon, line 444: it seems more logical/easier to read to `emit` `data.interimAsset`, because that is used in line 442
(note `r.interimToken` isn't updated because r is `memory`, although it has been set to the right value before)

src/router-plus/SuperformRouterPlusAsync.sol
```diff
@@ -0,0 +1,558 @@
...
+            vars.rebalanceToCallData
+        );
+
+        emit XChainRebalanceComplete(args_.receiverAddressSP, args_.routerPlusPayloadId);
+
+        return true;
+    }
+
+    /// @inheritdoc ISuperformRouterPlusAsync
+    function requestRefund(uint256 routerPlusPayloadId_, uint256 requestedAmount) external {
```
gpersoon, line 430: `requestRefund()` can be done multiple times
Assume it is done for a low amount and then it is approved via `approveRefund()`.
Then `approveRefund()` is front-run by a second `requestRefund()` which increases the `requestedAmount` to
a larger amount (although still `<= data.expectedAmountInterimAsset`)

Then the second amount will be approved

possible solutions: only allow `requestRefund()` once, or supply and check the amount in `approveRefund()`

src/router-plus/SuperformRouterPlusAsync.sol
```diff
@@ -0,0 +1,558 @@
...
+        refunds[routerPlusPayloadId_] = Refund(msg.sender, data.interimAsset, requestedAmount);
+
+        emit refundRequested(routerPlusPayloadId_, msg.sender, r.interimToken, requestedAmount);
+    }
+
+    /// @inheritdoc ISuperformRouterPlusAsync
+    function approveRefund(uint256 routerPlusPayloadId_) external onlyCoreStateRegistryRescuer {
+        if (approvedRefund[routerPlusPayloadId_]) revert REFUND_ALREADY_APPROVED();
+
+        Refund memory r = refunds[routerPlusPayloadId_];
```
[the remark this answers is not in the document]
TamaraRingas: https://github.com/superform-xyz/superform-core/pull/656

src/router-plus/SuperformRouterPlus.sol
TamaraRingas: https://github.com/superform-xyz/superform-core/pull/657

TamaraRingas (approved):

PR #619, opened 2024-09-06
chore: add decimal corrections for payment-helper

sujithsomraaj:


src/payments/PaymentHelper.sol
```diff
@@ -50,7 +50,8 @@ contract PaymentHelper is IPaymentHelper {
     //////////////////////////////////////////////////////////////
 
     uint256 private constant PROOF_LENGTH = 160;
-    uint8 private constant SUPPORTED_FEED_PRECISION = 8;
+    uint8 private constant MIN_FEED_PRECISION = 8;
+    uint8 private constant MAX_FEED_PRECISION = 8;
```
0xTimepunk, line 54: isn't this 18?

src/payments/PaymentHelper.sol
```diff
@@ -999,7 +1005,11 @@ contract PaymentHelper is IPaymentHelper {
             ) {
                 if (value <= 0) revert Error.CHAINLINK_MALFUNCTION();
                 if (updatedAt == 0) revert Error.CHAINLINK_INCOMPLETE_ROUND();
-                return uint256(value);
+
+                uint256 oraclePrecision = _getOracleDecimals(AggregatorV3Interface(oracleAddr));
+
+                if (oraclePrecision == MIN_FEED_PRECISION) return uint256(value);
+                else return uint256(value) / (10 ** oraclePrecision - MIN_FEED_PRECISION);
```
0xTimepunk, line 1012: Validate this calculation with eOracle. Also shouldn't we do 10 **(oraclePrecision - MIN_FEED_PRECISION) ?

0xTimepunk (commented):
Requires re-review

src/payments/PaymentHelper.sol
```diff
@@ -999,7 +1005,11 @@ contract PaymentHelper is IPaymentHelper {
             ) {
                 if (value <= 0) revert Error.CHAINLINK_MALFUNCTION();
                 if (updatedAt == 0) revert Error.CHAINLINK_INCOMPLETE_ROUND();
-                return uint256(value);
+
+                uint256 oraclePrecision = _getOracleDecimals(AggregatorV3Interface(oracleAddr));
```
gpersoon, line 1009: This piece of code retrieves the decimals every time from the oracle
If you want to save gas, this value could be combined with the oracle address and stored in one slot.
This does make the code more complex though

PR #626, opened 2024-09-20
chore: fix router plus no txdata bug

sujithsomraaj:


src/router-plus/SuperformRouterPlusAsync.sol
```diff
@@ -546,6 +546,8 @@ contract SuperformRouterPlusAsync is ISuperformRouterPlusAsync, BaseSuperformRou
         }
 
         for (uint256 i; i < sfDataLen; ++i) {
+            /// @dev if txData is empty, no update is made
+            if (liqRequests[i].txData.length == 0) continue;
```
0xTimepunk, line 550: with this if statement, we never enter the other if statement below: (sfData.liqRequests[i].token != address(0) && liqRequests[i].txData.length == 0)

Also, if txData length is 0 and token is 0 too we still need to update
```
sfData.amounts[i] = amounts[i];
sfData.outputAmounts[i] = outputAmounts[I];
```

0xTimepunk (changes requested):

src/router-plus/SuperformRouterPlusAsync.sol
sujithsomraaj: fixed

0xTimepunk (approved):

PR #631, opened 2024-10-01
feat: optimize router plus for backend needs

0xTimepunk:
This PR:
- Changes scripts to add a role
- Simplifies the way we pass ambIds and dstChains for Async Router Plus execution of rebalances
- Removes interimToken check and allows update by a keeper on Async Router Plus complete cross chain rebalance


subhasishgoswami (approved):
Looks good

PR #639, opened 2024-10-16
fix: Guard against malicious processors [SUP-8873]

TamaraRingas:
## Problem

A malicious processor can take advantage of the refund mechanism.

`completeCrossChainRebalance()` complete a cross chain rebalance initiated by the user with `startCrossChainRebalance()`. This function could be called only by an address with a `ROUTER_PLUS_PROCESSOR_ROLE`. As the processor can pass arbitrary data as function arguments to both functions he can take advantage of the refund mechanism, leading to two possible scenarios:

Processor can force unnecessary refunds in `completeCrossChainRebalance()`: by passing a specific `expectedAmountInterimAsset`, he can force unnecessary refunds on every payload to be processed.
Processor can steal all SuperformRouterPlusAsync funds: by starting a cross chain rebalance himself and thus passing a fake `expectedAmountInterimAsset`, he can issue a refund to himself stealing funds from the SuperformRouterPlusAsync contract.

## Solution

- Refactor refund mechanism to include a `requestRefund()` function and a corresponding `approveRefund()` function that is only callable by `CORE_STATE_REGISTRY_RESCUER`

- Update tests


src/router-plus/SuperformRouterPlusAsync.sol
```diff
@@ -415,42 +425,29 @@ contract SuperformRouterPlusAsync is ISuperformRouterPlusAsync, BaseSuperformRou
...
-    function proposeRefund(uint256 routerPlusPayloadId_, uint256 refundAmount_) external {
-        if (!_hasRole(keccak256("CORE_STATE_REGISTRY_RESCUER_ROLE"), msg.sender)) revert INVALID_PROPOSER();
+    function approveRefund(uint256 routerPlusPayloadId_) external onlyCoreStateRegistryRescuer {
+        if (approvedRefund[routerPlusPayloadId_]) revert REFUND_ALREADY_APPROVED();
 
-        Refund storage r = refunds[routerPlusPayloadId_];
-
-        if (r.interimToken == address(0) || r.receiver == address(0)) revert INVALID_REFUND_DATA();
-        if (r.proposedTime != 0) revert REFUND_ALREADY_PROPOSED();
+        Refund memory r = refunds[routerPlusPayloadId_];
```
0xTimepunk, line 444: Let's remove `proposedTime` from Refund struct in ISuperformRouterPlusAsync.sol

0xTimepunk (commented):
Changes in RouterPlus async look good to me.
Please review formatting changes in all the other src files and forge-scripts folder (which should be reverted to their original state).

Once you commit CI should run and let us know if there was any coverage drop with the function changes.

Also wondering why yAudit mentioned to revert at slippage check failure, it doesn't seem necessary to me - after we clean up let's bring this PR for discussion in discord for them to review the fix.

src/router-plus/SuperformRouterPlusAsync.sol
```diff
@@ -415,42 +425,30 @@ contract SuperformRouterPlusAsync is ISuperformRouterPlusAsync, BaseSuperformRou
...
-
-        emit RefundDisputed(routerPlusPayloadId_, msg.sender);
-    }
+    function requestRefund(uint256 routerPlusPayloadId_, uint256 requestedAmount) external {
+        Refund memory r = refunds[routerPlusPayloadId_];
 
-    /// @inheritdoc ISuperformRouterPlusAsync
-    function proposeRefund(uint256 routerPlusPayloadId_, uint256 refundAmount_) external {
-        if (!_hasRole(keccak256("CORE_STATE_REGISTRY_RESCUER_ROLE"), msg.sender)) revert INVALID_PROPOSER();
+        if (msg.sender != r.receiver) revert INVALID_REQUESTER();
```
0xTimepunk, line 431: Can add an extra check to see if refund already has been approved

src/router-plus/SuperformRouterPlusAsync.sol
```diff
@@ -253,8 +263,8 @@ contract SuperformRouterPlusAsync is ISuperformRouterPlusAsync, BaseSuperformRou
             ENTIRE_SLIPPAGE * args_.amountReceivedInterimAsset
                 < ((data.expectedAmountInterimAsset * (ENTIRE_SLIPPAGE - data.slippage)))
         ) {
-            refunds[args_.routerPlusPayloadId] =
-                Refund(args_.receiverAddressSP, data.interimAsset, args_.amountReceivedInterimAsset, block.timestamp);
+
+            refunds[args_.routerPlusPayloadId] = Refund(args_.receiverAddressSP, data.interimAsset, 0);
```
0xTimepunk, line 267: Why do we send amount as 0? Should just delete this from the structure entirely :)

src/router-plus/SuperformRouterPlusAsync.sol
```diff
@@ -415,42 +425,30 @@ contract SuperformRouterPlusAsync is ISuperformRouterPlusAsync, BaseSuperformRou
...
 
-        if (r.interimToken == address(0) || r.receiver == address(0)) revert INVALID_REFUND_DATA();
-        if (r.proposedTime != 0) revert REFUND_ALREADY_PROPOSED();
+        if (requestedAmount > data.expectedAmountInterimAsset) {
+            revert REQUESTED_AMOUNT_TOO_HIGH();
+        }
 
-        r.proposedTime = block.timestamp;
-        r.amount = refundAmount_;
+        refunds[routerPlusPayloadId_] = Refund(msg.sender, data.interimAsset, requestedAmount);
```
0xTimepunk, line 440: why do we write the whole structure again? msg.sender is still r.receiver, interim asset is the same asset too,

just probably need to change 
        r.amount = requestedAmount;

0xTimepunk (commented):
Few final fixes and LGTM

src/router-plus/SuperformRouterPlusAsync.sol
0xTimepunk: EDIT: maybe this is cleared because of the delete, but still

src/router-plus/SuperformRouterPlusAsync.sol
TamaraRingas: It was giving an error saying the argument count was off, I could do something like:

```
Refund r = new Refund(
   receiverAddressSP = args_.receiverAddressSP,
   asset =  data.interimAsset
);

and then say refunds[args_.routerPlusPayloadId] =  r
```

src/router-plus/SuperformRouterPlusAsync.sol
TamaraRingas: I thought we could just do `r.amount = requestedAmount` but this failed in the tests (r.amount) was not updating

src/router-plus/SuperformRouterPlusAsync.sol
0xTimepunk: I see, nevermind, can send as 0 I guess

0xTimepunk (approved):

PR #644, opened 2024-10-21
test: fix amount in calculation

0xTimepunk:
@TamaraRingas please understand the difference between callData and rebalanceToCalldata, where amountIn pertains to the interimAsset amount going in to the rebalanceTo step. Thereby we need to validate it by de-structuring the call data according to all possible DEPOSIT possibilities. It is a bit more complex and initially didn't foresee we'd need to do this, but I believe it solves the problem


TamaraRingas (approved):

TamaraRingas:
Ah I see, thanks for explaining this! LGTM 😁

PR #645, opened 2024-10-23
fix: insert tolerance into `redeemShare()` checks [SUP-8853]

TamaraRingas:
https://github.com/yAudit/superform-report/issues/2

deposit4626() is used to redeem vault assets and deposit them in a superform vault. However, in _redeemShare(), there is a strict equality balance check, that, in case of stETH, will always revert.

Technical Details

Lido stETH has a know rounding down problem as stated in their [docs](https://docs.lido.fi/guides/lido-tokens-integration-guide/#1-2-wei-corner-case). When trying to redeem stETH with [deposit4626()](https://github.com/superform-xyz/superform-core/blob/f7f06d763a6af148899e9c292f1a57414db60cc3/src/router-plus/SuperformRouterPlus.sol#L356-L390), the [balance check](https://github.com/superform-xyz/superform-core/blob/f7f06d763a6af148899e9c292f1a57414db60cc3/src/router-plus/SuperformRouterPlus.sol#L522) will always fail leading the function to revert.

Impact

Low. It is not possible to use deposit4626() with a stETH vault and any other token that have a similar behaviour.

Recommendation

Avoid to use strict equality and insert a tolerance in the balance check to handle the 1-2 wei corner case in the same way that is done [here](https://github.com/superform-xyz/superform-core/blob/f7f06d763a6af148899e9c292f1a57414db60cc3/src/forms/ERC5115Form.sol#L30-L31).

### DIFF
--- a/src/router-plus/SuperformRouterPlus.sol
+++ b/src/router-plus/SuperformRouterPlus.sol
@@ -521,11 +521,11 @@ contract SuperformRouterPlus is ISuperformRouterPlus, BaseSuperformRouterPlus {
         uint256 assetsBalanceAfter = asset.balanceOf(address(this));
         balanceDifference = assetsBalanceAfter - assetsBalanceBefore;
 
-        if (assets < TOLERANCE_CONSTANT) revert TOLERANCE_EXCEEDED();
+        if (assets < TOLERANCE_CONSTANT || balanceDifference < TOLERANCE_CONSTANT) revert TOLERANCE_EXCEEDED();
 
         /// @dev validate the slippage
         if (
-            (balanceDifference < assets - TOLERANCE_CONSTANT)
+            (balanceDifference != assets)
                 || (ENTIRE_SLIPPAGE * assets < ((expectedOutputAmount_ * (ENTIRE_SLIPPAGE - maxSlippage_))))
         ) {
             revert ASSETS_RECEIVED_OUT_OF_SLIPPAGE();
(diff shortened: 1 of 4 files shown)


src/router-plus/SuperformRouterPlus.sol
```diff
@@ -521,6 +524,7 @@ contract SuperformRouterPlus is ISuperformRouterPlus, BaseSuperformRouterPlus {
         if (
             (balanceDifference != assets)
                 || (ENTIRE_SLIPPAGE * assets < ((expectedOutputAmount_ * (ENTIRE_SLIPPAGE - maxSlippage_))))
+                || (assets - TOLERANCE_CONSTANT > balanceDifference)
```
0xTimepunk, line 527: Think this if statement is really big. Is it possible to use a new error (TOLERANCE_EXCEEDED)?

src/router-plus/SuperformRouterPlus.sol
```diff
@@ -27,6 +27,9 @@ contract SuperformRouterPlus is ISuperformRouterPlus, BaseSuperformRouterPlus {
 
     uint256 public ROUTER_PLUS_PAYLOAD_ID;
 
+    /// @dev Tolerance constant to account for tokens with rounding issues on transfer
+    uint256 constant TOLERANCE_CONSTANT = 2 wei;
```
0xTimepunk, line 31: Let's use 10 WEI just like in 5115

0xTimepunk (commented):
Almost there!

src/router-plus/SuperformRouterPlus.sol
TamaraRingas: Sure thing!

src/router-plus/SuperformRouterPlus.sol
TamaraRingas: Okay cool 👍

0xTimepunk (approved):

PR #657, opened 2024-11-22
fix: remedy rounding issue in `_deposit4626()` by adding remainder to `_deposit()` call

TamaraRingas:
`uint256 msgValue = msg.value / arrayLength;` rounds down so the last bit of `msg.value` is not used.

- Derive `valuePerItem`
- Get remainder from above
- Pass the sum of these values in the call to `deposit()`

### DIFF
--- a/src/router-plus/SuperformRouterPlus.sol
+++ b/src/router-plus/SuperformRouterPlus.sol
@@ -627,6 +627,7 @@ contract SuperformRouterPlus is ISuperformRouterPlus, BaseSuperformRouterPlus {
     /// @notice deposits ERC4626 vault shares into superform
     /// @param vault_ The ERC4626 vault to redeem from
     /// @param args Rest of the arguments to deposit 4626
+    /// @param arrayLength The length of the array of deposit4626 calls
     function _deposit4626(address vault_, Deposit4626Args calldata args, uint256 arrayLength) internal {
         _transferERC20In(IERC20(vault_), args.receiverAddressSP, args.amount);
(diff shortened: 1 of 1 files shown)


src/router-plus/SuperformRouterPlus.sol
```diff
@@ -639,10 +640,15 @@ contract SuperformRouterPlus is ISuperformRouterPlus, BaseSuperformRouterPlus {
 
         uint256 amountIn = _validateAndGetAmountIn(args.depositCallData, amountRedeemed);
 
-        uint256 msgValue = msg.value / arrayLength;
         address router = _getAddress(keccak256("SUPERFORM_ROUTER"));
+        uint256 valueToDeposit;
+        {
+            uint256 valuePerItem = (msg.value - (msg.value % arrayLength)) / arrayLength;
+            uint256 remainingValue = msg.value % arrayLength;
+            valueToDeposit = valuePerItem + remainingValue;
```
gpersoon, line 648: the `remainingValue` should only be used once  - for one of the vaults (or discounted in another way, or neglected)

this calculation: `uint256 valuePerItem = msg.value / arrayLength` can stay as it was because it rounds down anyway

Perhaps do the calculation of the `valueToDeposit` in the calling function `deposit4626()` and add the `remainingValue` only for the last (or first) item in the loop.

0xTimepunk: [2378603](https://github.com/superform-xyz/superform-core/pull/661/commits/237860315756e82f4a4bc25e2c2c138d7b359986)

